PR #202, opened 2025-10-15
ref(distributed_algorithms): Modular algorithm implementation

nicola-bastianello:
Implement new gradient-based algorithms: Adapt-Then-Combine, Aug-DGM, Wang-Elia


elramen:
Nice work! My comments:
1. Please split the PR into 2 separate: one for the modularization (ref) and one for the new algorithms (feat)
2. Why make the new methods private? They should be public I think
3. Start the commit message with a verb, and the scope can be shortened to dst_algs or just algs.

PR #204, opened 2025-10-15
feat(algs): Add new gradient-based algorithms

nicola-bastianello:
Add new gradient-based algorithms: Adapt-Then-Combine, Aug-DGM, Wang-Elia

### DIFF
--- a/decent_bench/distributed_algorithms.py
+++ b/decent_bench/distributed_algorithms.py
@@ -296,6 +296,8 @@ class AugDGM(DstAlgorithm):
             #     step 1: perform local gradient step and communicate
             for i in network.get_active_agents(k):
                 i.aux_vars["s"] = i.x - self.step_size * i.aux_vars["y"]
+
+            for i in network.get_active_agents(k):
                 network.broadcast(i, i.aux_vars["s"])
 
             #     step 2: update state and compute new local gradient


decent_bench/distributed_algorithms.py
```diff
@@ -178,6 +238,165 @@ def run(self, network: Network) -> None:
...
+
+    """
+
+    iterations: int
+    step_size: float
+    name: str = "Aug-DGM"
+
+    def run(self, network: Network) -> None:
+        r"""
+        Run the algorithm with all :math:`\mathbf{x}` and :math:`\mathbf{y}` initialized using :func:`numpy.zeros`.
```
elramen: Comment needs to be updated, y is not initialized to zeros. Should the other aux vars' initialization be mentioned here as well?

decent_bench/distributed_algorithms.py
```diff
@@ -71,6 +71,66 @@ def run(self, network: Network) -> None:
...
+            # transmit and receive
+            for i in network.get_active_agents(k):
+                network.broadcast(i, i.aux_vars["y"])
+
+            # consensus (a.k.a. combine step)
+            for i in network.get_active_agents(k):
+                network.receive_all(i)
+                neighborhood_avg = np.sum([W[i, j] * x_j for j, x_j in i.received_messages.items()], axis=0)
+                neighborhood_avg += W[i, i] * i.x
+                i.x = neighborhood_avg
```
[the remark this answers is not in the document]
nicola-bastianello: Good point, yes, let's split single "actions" as much as possible. The only issue is that people who implement their own algorithms might not do this, leading to unfair comparisons. But that's an issue for later

decent_bench/distributed_algorithms.py
nicola-bastianello: Well spotted, I copy-pasted and forgot to change. I think I would actually prefer to write only "Run the algorithm." in all of them. Later they will not be initialized to zero, so the sentence will become inaccurate at some point.

decent_bench/distributed_algorithms.py
```diff
@@ -178,6 +240,173 @@ def run(self, network: Network) -> None:
...
+            )
+
+        W = network.metropolis_weights  # noqa: N806
+
+        for k in range(self.iterations):
+            # 1st communication round
+            #     step 1: perform local gradient step and communicate
+            for i in network.get_active_agents(k):
+                i.aux_vars["s"] = i.x - self.step_size * i.aux_vars["y"]
+                network.broadcast(i, i.aux_vars["s"])
```
elramen: Should this update also be split from broadcast to keep all updates and communications separate?

decent_bench/distributed_algorithms.py
```diff
@@ -71,6 +71,68 @@ def run(self, network: Network) -> None:
...
+        for k in range(self.iterations):
+            # gradient step (a.k.a. adapt step)
+            for i in network.get_active_agents(k):
+                i.aux_vars["y"] = i.x - self.step_size * i.cost_function.gradient(i.x)
+
+            # transmit and receive
+            for i in network.get_active_agents(k):
+                network.broadcast(i, i.aux_vars["y"])
+
+            # consensus (a.k.a. combine step)
```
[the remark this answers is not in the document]
nicola-bastianello: No, in my view the combine step includes both reception and averaging

decent_bench/distributed_algorithms.py
nicola-bastianello: Sure. But if even I struggle with this, I can imagine users will too. We need to discuss a good way to accomplish this

elramen (approved):

PR #210, opened 2025-10-22
feat(algs): Add more gradient-based algorithms

nicola-bastianello:
Add algorithms: EXTRA, NIDS, ATC-Tracking, ADMM-Tracking Gradient (ATG)

### DIFF
--- a/decent_bench/distributed_algorithms.py
+++ b/decent_bench/distributed_algorithms.py
@@ -893,12 +893,12 @@ class DLM(DstAlgorithm):
                 )
 
             # step 2: communication round
-            for i in network.get_active_agents():
+            for i in network.get_active_agents(k):
                 network.broadcast(i, i.x)
-            for i in network.get_active_agents():
+            for i in network.get_active_agents(k):
                 network.receive_all(i)
             # compute and store \sum_j (\mathbf{x}_{i,k+1} - \mathbf{x}_{j,k+1})
-            for i in network.get_active_agents():
+            for i in network.get_active_agents(k):
                 i.aux_vars["s"] = np.sum([i.x - x_j for _, x_j in i.received_messages.items()], axis=0)
 
             # step 3: update dual variable


decent_bench/distributed_algorithms.py
```diff
@@ -409,6 +409,261 @@ def run(self, network: Network) -> None:
...
+        # first iteration (iteration k=1)
+        for i in network.get_active_agents(0):
+            network.broadcast(i, i.x)
+        for i in network.get_active_agents(0):
+            network.receive_all(i)
+        for i in network.get_active_agents(0):
+            neighborhood_avg = np.sum([W[i, j] * x_j for j, x_j in i.received_messages.items()], axis=0)
+            neighborhood_avg += W[i, i] * i.x
+            i.aux_vars["x_cons"] = neighborhood_avg  # store W x_k
+            i.aux_vars["x_old"] = i.x  # store x_0
```
elramen: [not blocking] should the second comment say x_0 instead of x_k?

decent_bench/distributed_algorithms.py
```diff
@@ -468,3 +723,187 @@ def run(self, network: Network) -> None:
...
+        for i in network.get_all_agents():
+            network.broadcast(i, i.x)
+        for i in network.get_all_agents():
+            network.receive_all(i)
+        # compute and store \sum_j (\mathbf{x}_{i,0} - \mathbf{x}_{j,0})
+        for i in network.get_all_agents():
+            i.aux_vars["s"] = np.sum([i.x - x_j for _, x_j in i.received_messages.items()], axis=0)
+
+        # main iteration
+        for k in range(self.iterations):
```
elramen: [blocking] unless I'm missing something, get_active_agents should be used instead of get_all_agents (and the main loop should use range(1, self.iterations)) like in the other new algorithms

nicola-bastianello: Totally right, I used autocomplete and thought it was get_active_agents

Issue #203, opened 2025-10-15
Allow custom consensus weights in Network object


nicola-bastianello:
Currently the default, which cannot be changed, is Metropolis weights. But the user might want to test alternatives. I propose the following changes:

- [ ] have a `weights` attribute in the `Network`, which by default is initialized to Metropolist weights
- [ ] spin off `metropolis_weights` into a standalone util

PR #223, opened 2025-12-03
feat(Agents): Generalize optimization parameter

### LINKED ISSUE #200: Generalize optimization parameter type beyond numpy array
### Problem
The optimization parameter ("X") in the codebase is currently expected to be a numpy array. This assumption limits users from utilizing alternative libraries and frameworks such as TensorFlow, PyTorch, JAX, and other linear algebra or machine learning frameworks that might use their own array types.

### Impact
This restriction significantly constrains the flexibility of the library, making it less attractive for researchers and practitioners who may prefer or require these other frameworks for advanced differentiation, GPU acceleration, or integration with their existing codebases. Generalizing the parameter type would make `decent-bench` much more extensible and user-friendly.

### Required Work
Changing the expectation from a numpy array to a more abstract or generic type will likely require a major refactor throughout the codebase:
- All functions and classes relying on numpy-specific methods will need to be audited
- Interface and type checks must be adapted or made more flexible
- Documentation and potential downstream examples/tests will need updating
- Backward compatibility considerations should be explicitly addressed

### Request for Input
Given the scale of this rework, input from maintainers and community is needed:
- Should we use protocols, duck typing, or something like Python's `__array__` interface?
- Is it worth supporting all major frameworks (TF, PyTorch, JAX) natively, or should we rely on user-side conversions?
- Any recommended approaches, patterns, or libraries to help abstract array-like behavior?

**This is a significant architectural change and help is needed to decide on the best approach.**

Please share your thoughts and suggestions on how to proceed!

Simpag:
Implements a new array type called `Array`, alongside some helper typealiases/enums for `SupportedArrayTypes`, `SupportedFrameworks` and `SupportedDevices`. 

* This new `Array` class make operations between different frameworks agnostic alongside the interoparability layer. No automatic casting between frameworks are performed due to performance. However, this should not be an issue as the only reason two different frameworks would be used is if a dataset is used with an incompatible Cost. 
* During typechecking the new `Array` class is used throughout Decent Bench while during runtime the `Array` class is unwrapped to its base framework for performance reasons.
* Modifications have been made to the interoperability layer so that it operates on `Array`. Some more functionality has been added and better modularization. 
* A custom decorator for cost functions is added to allow cost functions to operate directly on its underlying framework while allowing for type checking. During runtime, this decorator will unpack the `Array` object and convert it to the desired framework silently if needed. If conversion is needed it will throw a warning to tell the user that this behavior might cause performance issues. 
* Datasets have been updated to return `tuple[Array, Array]`.
* Network weights updated to use the new `Array` class, a new adjacency property which returns the adjacency matrix and a set_weights function is added which all should solve issue #203.
* Documentation updates to reflect these changes. 
* Minor fixes for issues created by updating sphinx 8.2.3 -> 9.0.1.
* Optional development environment with gpu support for tensorflow/jax. 

closes #200, #203, #224

### DIFF
--- a/docs/source/api/snippets/proximal_operator.rst
+++ b/docs/source/api/snippets/proximal_operator.rst
@@ -1,5 +1,5 @@
 .. math::
-    \operatorname{prox}_{\rho f}(\mathbf{y})
-    = \arg\min_{\mathbf{x}}  \left\{ f(\mathbf{x}) + \frac{1}{2\rho} \| \mathbf{x} - \mathbf{y} \|^2 \right\}
+    \operatorname{prox}_{\rho f}(\mathbf{x})
+    = \arg\min_{\mathbf{y}}  \left\{ f(\mathbf{y}) + \frac{1}{2\rho} \| \mathbf{y} - \mathbf{x} \|^2 \right\}
 
 where :math:`\rho > 0` is the penalty and :math:`f` the cost function.


docs/source/user.rst
```diff
@@ -350,6 +370,80 @@ Create your own metrics to tabulate and/or plot.
...
+
+        def __add__(self, other: Cost) -> Cost:
+            # Support addition of costs
+            if self.shape != other.shape:
+                raise ValueError(f"Mismatching domain shapes: {self.shape} vs {other.shape}")
+
+            return SumCost([self, other])
+
+
 Output
```
nicola-bastianello: I would move the "output" section, maybe inside the "sunny case" section

decent_bench/utils/array.py
```diff
@@ -0,0 +1,345 @@
...
+import decent_bench.utils.interoperability as iop
+from decent_bench.utils.types import ArrayKey, SupportedArrayTypes
+
+
+class Array:  # noqa: PLR0904
+    """
+    A wrapper class for :data:`decent_bench.utils.types.SupportedArrayTypes` objects to enable operator overloading.
+
+    This class allows for seamless interoperability between different tensor frameworks
+    by overloading standard arithmetic operators.It supports operations such as addition, subtraction, multiplication,
```
nicola-bastianello: space missing in ".. operators.It .."

decent_bench/utils/interoperability/_operators.py
```diff
@@ -0,0 +1,483 @@
...
+        return _return_array(torch.add(value1, value2))
+    if tf and isinstance(value1, _tf_types) and isinstance(value2, _tf_types):
+        return _return_array(tf.add(value1, value2))
+    if jnp and isinstance(value1, _jnp_types) and isinstance(value2, _jnp_types):
+        return _return_array(jnp.add(value1, value2))
+
+    raise TypeError(f"Unsupported framework type: {type(value1)} and {type(value2)}")
+
+
+def iadd[T: Array](array1: T, array2: Array | SupportedArrayTypes) -> T:
```
nicola-bastianello: this and other "i" operations use type `T`, which doesn't seem to be defined since the docs don't link to it. maybe a leftover from previous version?

Simpag: All of the "i" operators are inplace operators so they modify the value directly, therefore in order to keep type checking it indicates that the function returns the same instance that is inputted to the first argument. This type var "T" shows that the same instance is returned and sphinx autodocs are awful at documenting these. In order for them to be documented you need to write some very hacky code (to the best of my knowledge) which I thought was somewhat unnecessary since these "i" operators should not really be used by users, instead they should use the normal operators which returns a new instance. These "i" operators are used internally for the __iadd__ etc dunder methods of the `Array` class. The docstring does mention that the input argument array1 is of type Array so I think doc readers should be able to infer what type it is but I can add the hacky solution if we really want these type vars to be documented. 

What I also could do is that I remove the "i" operators from the public API so that they are not included in the docs nor available to users but if a user really knows what they are doing it can be slightly more performant to user iadd in some cases.

docs/source/api/decent_bench.utils.array.rst
```diff
@@ -0,0 +1,7 @@
+decent\_bench.utils.array
```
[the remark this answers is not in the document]
Simpag: I think its best if we dont doc any of the dunder methods and just write a better doc string for the class.

decent_bench/utils/interoperability/_functions.py
```diff
@@ -0,0 +1,928 @@
...
+        return dtype(value.item())
+    if tf and isinstance(value, tf.Tensor):
+        return dtype(to_numpy(value).item())
+    if jnp and isinstance(value, _jnp_types):
+        return dtype(value.item())
+
+    raise TypeError(f"Unsupported type: {type(value)}")
+
+
+def norm(array: Array, p: float = 2) -> Array:
```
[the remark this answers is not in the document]
nicola-bastianello: I see, thank you for the explanation

decent_bench/utils/interoperability/_decorators.py
```diff
@@ -0,0 +1,108 @@
...
+    It automatically converts input :class:`~decent_bench.utils.array.Array` arguments
+    to the cost's framework-specific array type and wraps the output based on the
+    superclass method's return type annotation.
+
+    Args:
+        superclass_method: The method from the superclass (e.g., `Cost.function`) that is being overridden.
+
+    Note:
+        Does not work on __add__ or similar special methods.
+        Only works when first argument is ``x`` or ``y``.
```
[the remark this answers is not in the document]
nicola-bastianello: also, I didn't catch this before, but I don't like having `y` as the argument of the proximal; in my opinion it should be `x`, so that the same argument name is used everywhere. I know it's a pain, but could you please also change this?

PR #226, opened 2025-12-10
fix(interoperability): stabilize tf pow precision

### LINKED ISSUE #225: fix TF float32 pow/ipow precision on Py3.13/Windows
On Windows with Python 3.13 and TensorFlow 2.20.0 (the only available wheel), float32 pow/ipow returns 81.00001 for 3**4, causing pow/ipow tests to fail (assert_almost_equal at decimal=6).

Environment
- OS: Windows
- Python: 3.13.9
- TensorFlow: 2.20.0 (only wheel for Py3.13 on Windows)
- Repo: current main (commit 581615d)

adrianardv:
- Cast TF float tensors to float64 for pow/ipow then back to avoid float32 ULP drift on TF 2.20/Py3.13.

- Add autodoc type aliases and ignore TypeAliasForwardRef to keep Sphinx nitpicky builds passing.

- closes #225

### DIFF
--- a/docs/source/conf.py
+++ b/docs/source/conf.py
@@ -30,10 +30,7 @@ extensions = [
 autodoc_default_options = {}
 autodoc_member_order = "bysource"
 autodoc_preserve_defaults = True
-autodoc_type_aliases = {
-    "SupportedArrayTypes": "decent_bench.utils.types.SupportedArrayTypes",
-    "ArrayKey": "decent_bench.utils.types.ArrayKey",
-}
+autodoc_type_aliases = {}
 
 nitpicky = True
 nitpick_ignore = [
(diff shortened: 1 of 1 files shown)


decent_bench/utils/interoperability/_operators.py
```diff
@@ -390,8 +396,13 @@ def ipow[T: Array](array: T, p: float) -> T:
...
         return cast("T", _return_array(value))
     if tf and isinstance(value, tf.Tensor):
+        orig_dtype = value.dtype
+        if getattr(orig_dtype, "is_floating", False):
+            result = tf.pow(tf.cast(value, tf.float64), p)
+            result = tf.cast(result, orig_dtype)
+            return cast("T", Array(result))
         value **= p
-        return cast("T", _return_array(value))
+        return cast("T", Array(value))
```
Simpag: Should always use the `_return_array()` function as this unwraps the `Array` class during runtime for performance reasons

nicola-bastianello:
so, do I understand correctly that this is an issue that pops up during the tests, but does not impact functionality otherwise? if that is the case then I would definitely only change the test. but if there are other elements of this PR that are also needed please let me know

adrianardv:
I added the ignores because my tox -e sphinx env was still on Sphinx 8.2.3 and, in nitpicky mode, it couldn’t resolve the SupportedArrayTypes/ArrayKey aliases, so the build failed. I hadn’t noticed the sphinx env was pinned to 8.2.3. After recreating it with Sphinx 9.0.4 (the version you’re using), the stock conf.py builds cleanly, so I’ll remove the ignores.

Simpag:
I see, thats good to know! @nicola-bastianello I have no other comments, LGTM

PR #229, opened 2025-12-12
feat(networks): add Network base class and FL network support

### LINKED ISSUE #192: Merge receive and receive_all
Currently `Network` implements `receive`, which allows to define the agent receiving and the sender (both as mandatory arguments); and `receive_all` with only receiver specified.

In my opinion, it would be better to have a single `receive` method with mandatory `receiver` argument, and optional `sender` argument. The latter can be:

- `None`, defaulting to receiving all available messages
- `Agent`, so a specific message is received
- list of `Agent` to specify a subset of messages to receive

The reasons to do this in my opinion:

1. easier coding of new algorithms: users just need to know to use `receive` (and `send` and `broadcast`)
2. it aligns well with federated scenarios where we only call `receive`; thus removing a scenario-specific method
3. increased flexibility by allowing lists of senders instead of one/all options currently available

adrianardv:
This PR extends the `networks` module to support a federated learning setup in addition to the existing peer-to-peer network.

- Introduces a common `Network` base class with shared communication attributes.
- Refactors `P2PNetwork` to inherit from `Network`.
- Adds an FL-style network that reuses the same interface and abstractions.
- Closes #192

### DIFF
--- a/decent_bench/networks.py
+++ b/decent_bench/networks.py
@@ -1,5 +1,5 @@
 from abc import ABC
-from collections.abc import Iterable, Mapping
+from collections.abc import Mapping, Sequence
 from functools import cached_property
 from typing import TYPE_CHECKING
 
@@ -83,7 +83,7 @@ class Network(ABC):  # noqa: B024
     def send(
         self,
         sender: Agent,
-        receiver: Agent | Iterable[Agent] | None = None,
+        receiver: Agent | Sequence[Agent] | None = None,
         msg: Array | None = None,
     ) -> None:
         """
@@ -91,7 +91,7 @@ class Network(ABC):  # noqa: B024
 
         Args:
             sender: sender agent
-            receiver: receiver agent, iterable of receiver agents, or ``None`` to broadcast to connected agents.
+            receiver: receiver agent, sequence of receiver agents, or ``None`` to broadcast to connected agents.
             msg: message to send
 
         Raises:
@@ -105,19 +105,20 @@ class Network(ABC):  # noqa: B024
         if sender not in self.graph:
             raise ValueError("Sender must be an agent in the network")
 
-        receivers: Iterable[Agent] | list[Agent]
         if receiver is None:
-            receivers = self.connected_agents(sender)
+            receiver = self.connected_agents(sender)
         elif isinstance(receiver, Agent):
-            receivers = [receiver]
-        else:
-            receivers = receiver
+            if receiver not in self.connected_agents(sender):
+                raise ValueError("Sender and receiver must be connected in the network")
+            self._send_one(sender=sender, receiver=receiver, msg=msg)
+            return
+        neighbors = set(self.connected_agents(sender))
+        invalid_receivers = [r for r in receiver if r not in neighbors]
+        if invalid_receivers:
+            ids = [r.id for r in invalid_receivers]
+            raise ValueError(f"Sender and receiver must be connected in the network; not connected receivers: {ids}")
 
-        receivers = list(receivers)
-        if any(r not in self.connected_agents(sender) for r in receivers):
-            raise ValueError("Sender and receiver must be connected in the network")
-
-        for r in receivers:
+        for r in receiver:
             self._send_one(sender=sender, receiver=r, msg=msg)
 
     def _receive_one(self, receiver: Agent, sender: Agent) -> None:
@@ -133,13 +134,13 @@ class Network(ABC):  # noqa: B024
             receiver._received_messages[sender] = msg  # noqa: SLF001
             self.graph.edges[sender, receiver][str(receiver.id)] = None
 
-    def receive(self, receiver: Agent, sender: Agent | Iterable[Agent] | None = None) -> None:
+    def receive(self, receiver: Agent, sender: Agent | Sequence[Agent] | None = None) -> None:
         """
         Receive message(s) at an agent.
 
         Args:
             receiver: receiver agent
-            sender: sender agent, iterable of sender agents, or ``None`` to receive from all connected agents.
+            sender: sender agent, sequence of sender agents, or ``None`` to receive from all connected agents.
 
         Raises:
             ValueError: if sender/receiver are not part of the network or not connected.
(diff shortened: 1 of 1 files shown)


decent_bench/networks.py
```diff
@@ -16,31 +19,115 @@
...
+
+    @property
+    def message_compression(self) -> CompressionScheme:
+        """Compression scheme applied to messages."""
+        return self._message_compression
+
+    @property
+    def message_drop(self) -> DropScheme:
+        """Drop scheme applied to messages."""
+        return self._message_drop
```
Simpag: What is the purpose of the properties graph, message_noise, message_compression and message_drop? I am not sure these would be used anywhere?

decent_bench/networks.py
```diff
@@ -149,31 +203,153 @@ def broadcast(self, sender: Agent, msg: Array) -> None:
...
 
         Received messages are stored in
         :attr:`Agent.messages <decent_bench.agents.Agent.messages>`.
         """
-        msg = self._graph.edges[sender, receiver].get(str(receiver.id))
-        if msg is not None:
-            receiver._n_received_messages += 1  # noqa: SLF001
-            receiver._received_messages[sender] = msg  # noqa: SLF001
-            self._graph.edges[sender, receiver][str(receiver.id)] = None
+        for neighbor in self.neighbors(receiver):
```
Simpag: Same as for broadcast

decent_bench/networks.py
```diff
@@ -149,31 +203,153 @@ def broadcast(self, sender: Agent, msg: Array) -> None:
...
+        """
+        clients = set(self.clients)
+        invalid = [client for client in msgs if client not in clients]
+        if invalid:
+            raise ValueError("All senders must be clients")
+        missing = clients - set(msgs)
+        if missing:
+            raise ValueError("Messages must be provided for all clients")
+        for client, msg in msgs.items():
+            self.send_from_client(client, msg)
```
Simpag: You could optimize this a bit by removing the list comprehension
```python 
def send_from_all_clients(self, msgs: Mapping[Agent, Array]) -> None:
    clients = set(self.clients)
    senders = set(msgs)

    # Check for invalid senders (those not in the client set)
    invalid = senders - clients
    if invalid:
        raise ValueError(f"Invalid senders: {invalid}")

    # Check for missing clients (those not in msgs)
    missing = clients - senders
    if missing:
        raise ValueError(f"Missing messages from: {missing}")

    for client, msg in msgs.items():
        self.send_from_client(client, msg)
```

decent_bench/networks.py
```diff
@@ -239,33 +254,133 @@ def _identify_server(self) -> Agent:
...
+                    raise ValueError("Server-to-server communication is not supported")
+                super().send(sender=sender, receiver=receiver, msg=msg)  # server -> client
+                return
+
+            if receiver is not self.server:
+                raise ValueError("Client-to-client communication is not supported")
+            super().send(sender=sender, receiver=receiver, msg=msg)  # client -> server
+            return
+
+        receivers = list(receiver)
```
Simpag: If you change from Iterable to Sequence you wont have to make a list here since you wont have to worry about maybe exhausting the iterator on line 338. Since you check above that its not an Agent instance, receiver will always be a list of agents

decent_bench/networks.py
```diff
@@ -239,33 +254,133 @@ def _identify_server(self) -> Agent:
...
+                    raise ValueError("Server-to-server communication is not supported")
+                super().receive(receiver=receiver, sender=sender)  # server receives from a client
+                return
+
+            if sender is not self.server:
+                raise ValueError("Client-to-client communication is not supported")
+            super().receive(receiver=receiver, sender=sender)  # client receives from the server
+            return
+
+        senders = list(sender)
```
Simpag: same as for send

decent_bench/networks.py
```diff
@@ -328,7 +442,7 @@ def receive_at_client(self, client: Agent) -> None:
         """
         if client not in self.clients:
             raise ValueError("Receiver must be a client")
-        self.receive(receiver=client, sender=self.server)
+        self.receive(receiver=client, sender=None)  # or self.receive(receiver=client, sender=self.server)
```
Simpag: Remove commented code

decent_bench/networks.py
adrianardv: Thanks for the idea! I tried tightening everything to Sequence so we could drop the list(...), but mypy didn’t like it: narrowing the override vs the base Iterable signature and some list/sequence assignment issues in the base blew up, plus a few other mypy errors popped up. Given that, I left the API as Iterable and kept the one materialisation here. That keeps typing happy and avoids iterator exhaustion. Happy to revisit if you have other suggestions.

decent_bench/networks.py
```diff
@@ -149,31 +203,154 @@ def broadcast(self, sender: Agent, msg: Array) -> None:
...
+        """Concrete network class for type-based dispatch."""
+        return type(self)
+
+    @property
+    def server(self) -> Agent:
+        """Agent acting as the central server."""
+        return self._server
+
+    @property
+    def clients(self) -> list[Agent]:
```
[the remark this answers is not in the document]
nicola-bastianello: This is great, exactly what I had in mind

decent_bench/networks.py
```diff
@@ -100,80 +218,254 @@ def adjacency(self) -> Array:
...
+            ValueError: if msg is missing or if sender/receiver roles are invalid.
 
-        Whether an :class:`~decent_bench.agents.Agent` is active or not at a given time is defined by its
-        :class:`~decent_bench.schemes.AgentActivationScheme`.
         """
-        return [a for a in self.agents() if a._activation.is_active(iteration)]  # noqa: SLF001
+        if msg is None:
+            raise ValueError("msg must be provided")
+
+        if sender not in self.graph:
```
nicola-bastianello: I think it's a great idea to check if the sender/receiver are indeed in the network, since it allows us to provide a customized error message. I realize that otherwise the error message popping up comes from the NetworkX backend and I don't know how informative that is.

Since this is a good thing to do, how about making it part of the superclass methods? This way both P2P and Fed inherit this check and we only have one place where the error message is specified

decent_bench/networks.py
```diff
@@ -16,29 +18,145 @@
...
+    def active_agents(self, iteration: int) -> list[Agent]:
+        """
+        Get all active agents.
+
+        Whether an :class:`~decent_bench.agents.Agent` is active or not at a given time is defined by its
+        :class:`~decent_bench.schemes.AgentActivationScheme`.
+        """
+        return [a for a in self.agents() if a._activation.is_active(iteration)]  # noqa: SLF001
+
+    def _adjacent_agents(self, agent: Agent) -> list[Agent]:
```
nicola-bastianello: Now I'm debating whether this should be a public method that is agnostic of the network type, just providing a list of the agents connected to any agent. And then `neighbors` is an alias, and the fed network also inherits the method; possibly using the method could simplify some of checks in send and receive of the fed network (e.g.: in a fed network we check at init that adjacent_agents of the server includes all clients, and adjacent_agents of any client is only the server; but then given a new send/receive operation we can do e.g. `if receiver not in seld.adjacent_agents(sender): ...`

Also: "adjacent" is widely used in p2p networks but less in fed ones; what if we called this "connected_agents"?

What do you think @adrianardv @Simpag ?

docs/source/user.rst
```diff
@@ -43,6 +43,16 @@ Benchmark executions will have outputs like these:
           :align: center
 
 
+Benchmark executions will have outputs like these:
```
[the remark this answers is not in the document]
adrianardv: Solved

decent_bench/networks.py
adrianardv: Good idea. I was looking at it, and I finally changed _adjacent_agents to a public connected_agents.
Network.send/receive now enforce that sender/receiver are in the network and connected, so the star topology validation at FedNetwork init, plus the base methods already make server-client comms work. In fact, both P2P and Fed would now correctly function with just the canonical send/receive.
Even so, I kept the Fed overrides only to surface FL-specific errors (no server-server or client-client, and only the server can fan outin to multiple peers). 
Let me know if you'd prefer dropping those overrides entirely and just relying on the base errors.

Simpag: Sounds good to me

PR #240, opened 2026-01-20
feat(Empirical): Empirical risk cost

### LINKED ISSUE #141: Batching in cost functions
Cost functions should support batching to enable stochastic algorithms. To discuss: should batch size be set in the constructor or be a method parameter (making it configurable in each evaluate/gradient/hessian/proximal call).

Note that m_smooth and m_cvx need to be updated.

Simpag:
This PR adds an empirical risk cost function and minor refactoring of the monolithic costs.py file. 

Empirical risk cost function implements support for stochastic gradients and batch sampling. Linear- and logistic regression costs have been converted to empirical risk cost functions. Setting batch-size to None will yield the same results as previously.

closes #141

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   mypy:
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
 
     steps:
       - name: Check out code
@@ -78,7 +78,7 @@ jobs:
         run: tox -e pytest
 
   ruff:
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
 
     steps:
       - name: Check out code
@@ -96,7 +96,7 @@ jobs:
         run: tox -e ruff
 
   sphinx:
-    runs-on: macos-latest
+    runs-on: ubuntu-latest
 
     steps:
       - name: Check out code


decent_bench/costs/_empirical_risk/_empirical_risk_cost.py
```diff
@@ -0,0 +1,130 @@
...
+from typing import Any
+
+import numpy as np
+
+from decent_bench.costs._base._cost import Cost
+from decent_bench.utils.array import Array
+
+
+class EmpiricalRiskCost(Cost, ABC):
+    """Base class for empirical risk cost functions."""
```
nicola-bastianello: it would be good to provide a more extensive docstring, where we show the mathematical definition of empirical risk cost, like eq (2) in https://arxiv.org/abs/2405.11999 (but without index i)

decent_bench/costs/_empirical_risk/_linear_regression_cost.py
```diff
@@ -0,0 +1,220 @@
...
+        return float(0.5 * residual.dot(residual))
+
+    @iop.autodecorate_cost_method(EmpiricalRiskCost.gradient)
+    def gradient(self, x: NDArray[float64], indices: list[int] | None = None) -> NDArray[float64]:
+        r"""
+        Gradient at x using datapoints at the given indices.
+
+        If indices is None, a random batch is drawn with :attr:`batch_size` samples.
+
+        .. math:: \mathbf{A}^T\mathbf{Ax} - \mathbf{A}^T \mathbf{b}
```
nicola-bastianello: same as for comment in line 134

decent_bench/costs/_empirical_risk/_linear_regression_cost.py
```diff
@@ -0,0 +1,220 @@
...
+        return res
+
+    @iop.autodecorate_cost_method(EmpiricalRiskCost.hessian)
+    def hessian(self, x: NDArray[float64], indices: list[int] | None = None) -> NDArray[float64]:  # noqa: ARG002
+        r"""
+        Hessian at x using datapoints at the given indices.
+
+        If indices is None, a random batch is drawn with :attr:`batch_size` samples.
+
+        .. math:: \mathbf{A}^T\mathbf{A}
```
nicola-bastianello: same as for comment in line 134

decent_bench/costs/_empirical_risk/_logistic_regression_cost.py
```diff
@@ -0,0 +1,210 @@
...
+from decent_bench.utils.types import SupportedDevices, SupportedFrameworks
+
+from ._empirical_risk_cost import EmpiricalRiskCost
+
+
+class LogisticRegressionCost(EmpiricalRiskCost):
+    r"""
+    Logistic regression cost function.
+
+    .. math:: f(\mathbf{x}) =
```
nicola-bastianello: same as linear regression cost, I would have both the current definition and the empirical risk definition of the cost

decent_bench/costs/_empirical_risk/_logistic_regression_cost.py
```diff
@@ -0,0 +1,210 @@
...
+        sig = special.expit(A.dot(x))
+        D = np.diag(sig * (1 - sig))  # noqa: N806
+        res: NDArray[float64] = A.T.dot(D).dot(A)
+        return res
+
+    def proximal(self, x: Array, rho: float) -> Array:  # type: ignore[override]
+        """
+        Proximal at x solved using an iterative method.
+
+        Note:
```
nicola-bastianello: I'm now debating whether proximal should never have support for indices, not even in linear regression (which is essentially the only empirical risk cost which has a closed form proximal). the idea could be: 1) in linear regression we always return the full proximal, ignoring batch size; 2) in any other empirical risk cost, we put a note saying that if there is no closed form solution, the proximal will be approximated using stochastic gradients. what do you think?

Simpag: I've been thinking about the proximal for a while. I was wondering if there's any point in implementing an approximation for the proximal at all? If there is no closed form solution to the proximal then the approximation strategy for solving this will affect the performance of an algorithm using the proximal, thus makes it somewhat ambiguous to assess performance. I personally think that it should be up to the algorithm to implement this approximation so that the performance is based on how well the algorithm performs, not based on our approximation of the proximal.

Simpag:
You understood the behavior correctly and I agree that this could be better documented. Currently this is only documented in the base class properties. 

Regarding your questions, we can add the option for indices to be a string literal but this will complicate things a bit. I have a solution to this which might make the base class a bit messier but will make it easier for users to create new costs. 

I cant really see a reason why someone would want to change the batch-size. Usually batch size is decided based on available resources so having that being scheduled in some way seems unlikely. But we can always add it if you think it would be of use

The reason why I set the batch_size to allow for None in init is because if we dont allow it to be None we cant have batch_size as a default argument since we cannot know the size of the dataset. We can of course set batch_size = n_samples during runtime in the init. I had that originally but I changed it to None to make the `_sample_batch_indices` method slightly faster but the improvement is miniscule so it shouldnt matter much.

decent_bench/costs/_empirical_risk/_linear_regression_cost.py
```diff
@@ -0,0 +1,220 @@
...
+from decent_bench.costs._empirical_risk._empirical_risk_cost import EmpiricalRiskCost
+from decent_bench.utils.array import Array
+from decent_bench.utils.types import SupportedDevices, SupportedFrameworks
+
+
+class LinearRegressionCost(EmpiricalRiskCost):
+    r"""
+    Linear regression cost function.
+
+    .. math:: f(\mathbf{x}) = \frac{1}{2} \| \mathbf{Ax} - \mathbf{b} \|^2
```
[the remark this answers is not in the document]
Simpag: Should it be like you wrote where we sum over every datapoint or should it be `\frac{1}{2} \sum_{i \in B} (A_i x - b_i)^2` where we sum over a batch?

decent_bench/costs/_empirical_risk/_empirical_risk_cost.py
nicola-bastianello: yes this looks great! just some comments on notation: 1) "n" is often the size of x, I see "m" being used more often as the number of total samples; 2) I would use "\ell" instead of "\mathcal{l}" for the loss

decent_bench/costs/_empirical_risk/_empirical_risk_cost.py
```diff
@@ -0,0 +1,130 @@
...
+            raise ValueError("No batch has been used yet.")
+        return self._last_batch_used
+
+    @cached_property
+    def _rand(self) -> np.random.Generator:
+        return np.random.default_rng(seed=0)  # Later replace with global rng
+
+    @abstractmethod
+    def predict(self, x: Array, data: list[Array]) -> Array:
+        """
```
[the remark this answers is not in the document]
nicola-bastianello: sounds good!

nicola-bastianello:
> You understood the behavior correctly and I agree that this could be better documented. Currently this is only documented in the base class properties.

great, thank you

> Regarding your questions, we can add the option for indices to be a string literal but this will complicate things a bit. I have a solution to this which might make the base class a bit messier but will make it easier for users to create new costs.

ok, then let's see the solution you propose; if we think it's too messy we can also avoid having the string literal option

> I cant really see a reason why someone would want to change the batch-size. Usually batch size is decided based on available resources so having that being scheduled in some way seems unlikely. But we can always add it if you think it would be of use

good point; then I agree it's not needed. in any case if an algorithm needs to change the batch size, it can do so by passing the `indices` directly which ignores the batch_size property (right?)

> The reason why I set the batch_size to allow for None in init is because if we dont allow it to be None we cant have batch_size as a default argument since we cannot know the size of the dataset. We can of course set batch_size = n_samples during runtime in the init. I had that originally but I changed it to None to make the `_sample_batch_indices` method slightly faster but the improvement is miniscule so it shouldnt matter much.

ok, then if the performance improvement is small, I would prefer to set batch_size = n_samples at runtime. Feels like the solution that will be easier to work with in the future (and if a user accesses the property, I think it's more expected to see an int instead of None)

Simpag:
Sphinx seems to fail because it randomly decided to use version `v8.2.3` instead of `v9.0.4` which it used earlier and we use locally? Maybe we can wait a bit and rerun the test. Also noticed that we run mypy and sphinx on macos, we might want to switch it to linux as it is most likely the most stable and up-to-date version.

Edit:
If I re-install my sphinx environment pip installs sphinx 8.2.3 for some reason even if sphinx 9.1.0 is the latest version.

Simpag:
It seems like we are always building from their latest main branch and not their stable releases. Let me update the config file to use the stable releases and hope that the checks use the one in the PR.

PR #243, opened 2026-02-05
feat(costs): Add dunder methods to costs

### LINKED ISSUE #178: Enhancing cost_functions
Add other magic methods to CostFunction:

- multiplication/division by a number (and only a number): `__mul__`, `__rmul__`, `__truediv__`, `__rtruediv__`
- handle reverse addition (useful when applying sum to a list of costs): `__radd__`
- negation `__neg__` (multiply by -1) (can be useful for creating non-convex costs)
- subtraction `__sub__` (sum by the negation of the other cost); also useful for nonconvex cases

adrianardv:
## Summary
- Add scalar multiplication, division, negation, subtraction, and reverse add for Cost
- Introduce ScaledCost to represent weighted objectives
- Add operator tests for the new behaviours

Closes #178

### DIFF
--- a/decent_bench/costs/_base/_scaled_cost.py
+++ b/decent_bench/costs/_base/_scaled_cost.py
@@ -60,6 +60,10 @@ class ScaledCost(Cost):
         return self.cost.hessian(x, *args, **kwargs) * self.scalar
 
     def proximal(self, x: Array, rho: float, *args: Any, **kwargs: Any) -> Array:  # noqa: ANN401
+        if rho <= 0:
+            raise ValueError("The penalty parameter rho must be positive.")
+        if self.scalar < 0:
+            raise ValueError("The proximal operator is not defined for negative scaling.")
         if self.scalar == 0:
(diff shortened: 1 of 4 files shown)


decent_bench/costs/_base/_scaled_cost.py
```diff
@@ -0,0 +1,70 @@
...
+    """A scalar multiple of another cost function."""
+
+    def __init__(self, cost: Cost, scalar: float):
+        self.cost: Cost
+        self.scalar: float
+        if isinstance(cost, ScaledCost):
+            self.cost = cost.cost
+            self.scalar = scalar * cost.scalar
+        else:
+            self.cost = cost
```
nicola-bastianello: the way we assign `self.cost` here is as a reference to `cost`, since that's the default in python. I'm thinking that maybe we should do a deepcopy to avoid unexpected behavior

@Simpag would there be any problem making deep copies of pytorch costs?

PR #255, opened 2026-03-04
enh(Agent): AgentHistory object for easier access to historical states

Simpag:
This PR implements a AgentHistory object which is a lightweight wrapper around a dictionary to make certain operations easier. This PR also includes a fix to a bug where the use of runtime metrics increased agent call counters. This is fixed by introducing a static agent function/context manager `Agent.no_count` that disables call counting for one or more agents.

### DIFF
--- a/decent_bench/agents.py
+++ b/decent_bench/agents.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import bisect
 import contextlib
-from collections.abc import Generator, Iterator, Mapping, Sequence
+from collections.abc import Iterator, Mapping, Sequence
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


decent_bench/agents.py
```diff
@@ -34,6 +36,7 @@ def __init__(self, agent_id: int, cost: Cost, activation: AgentActivationScheme,
         self._n_gradient_calls: float = 0
         self._n_hessian_calls: float = 0
         self._n_proximal_calls: float = 0
+        self._no_count_depth: int = 0  # Nesting counter; counting disabled when > 0
```
nicola-bastianello: why is this an int instead of a bool?

PR #261, opened 2026-03-14
Fix: algorithm instantiation and framework-specific interoperability bugs

adrianardv:
This PR fixes two follow-up issues I found after the recent iterations refactor.

First, when I ran the FedAvg sanity script, `FedAvg` could not be instantiated. The cause was that using `iterations: int = field()`  together with the abstract iterations property in ` Algorithm`  left the concrete algorithm classes abstract at runtime.

Second, when I ran ` pytest test`  in my local dev environment instead of only ` tox -e pytest` , I found additional framework-specific interoperability regressions in the PyTorch, TensorFlow, and JAX branches of ` maximum()`  and ` diag()` .

**Changes:**
- remove the abstract ` iterations`  property from ` Algorithm` 
- keep ` iterations`  documented in ` Algorithm`  as part of the common algorithm API
- add a test that instantiates distributed algorithms
- fix framework-specific interoperability bugs in: `maximum()` for PyTorch/TensorFlow/JAX; `diag()` forTensorFlow rank handling; JAX array type detection

**Verification:**
- tox
- pytest test
- fedavg_sanity.py (local script)

**Notes:**
On the iterations point, the current solution keeps the field order we wanted, without forcing keyword-only hyperparameters, avoids the dataclass non-default argument follows default argument issue, and removes the runtime conflict with the abstract iterations property. The change to Algorithm is small in practice because that abstract property was not being used as real runtime logic either way, and I still left a clear trace of iterations in Algorithm as a shared documented attribute.

### DIFF
--- a/decent_bench/distributed_algorithms.py
+++ b/decent_bench/distributed_algorithms.py
@@ -2,7 +2,7 @@ import random
 from abc import ABC, abstractmethod
 from collections.abc import Callable, Iterable, Sequence
 from dataclasses import dataclass, field
-from typing import TYPE_CHECKING, Final, cast, final
+from typing import TYPE_CHECKING, Any, Final, cast, final
 
 import decent_bench.utils.algorithm_helpers as alg_helpers
 import decent_bench.utils.interoperability as iop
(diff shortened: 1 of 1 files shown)


nicola-bastianello:
This is great! Both the iterations as non-abstract and testing instantiation of algorithms are great ideas

I'll merge right after #258 .

For the pytest issue: do you think we should update the tox config to make sure the proper tests are run?

adrianardv:
> This is great! Both the iterations as non-abstract and testing instantiation of algorithms are great ideas
> 
> I'll merge right after #258 .
> 
> For the pytest issue: do you think we should update the tox config to make sure the proper tests are run?

Yes, I think it would make sense to update the tox config so that it also checks the framework-specific branches.
As I see it, we have two options:
- update the tox pytest env so that tox/tox -e pytest also run the framework-specific tests.
- leave the tox config as it is, and note down somewhere that to check all framework branches people should run both tox and pytest test in an env where the optional framework dependencies are installed. In that case, CI would need an extra step/job to run those tests.

I think I would lean more towards option 1, just because it keeps everything inside the normal test workflow and avoids having to remember an extra step.

PR #274, opened 2026-03-27
feat(RNG): Cross-framework rng

### LINKED ISSUE #237: Setting a seed for random generators
Some users might be interested in setting a seed for random generators (numpy's, pytorch's, etc). Since I think all random generators allow to set a seed, we could extend the kwargs of interoperability random functions to allow for the seed.

Simpag:
This PR introduces a global rng system which can be used to set the seed (`iop.set_seed()`) for all used frameworks for the benchmark to be run. You set the seed before you generate any random datasets and the benchmark function. Results are reproducible with or without multiprocessing. Resumption is also equivalent to running the entire benchmark without interruption. Adding new iterations and/or trials is also equivalent to running the experiment from scratch, if a seed was set when the experiment started. Each algorithm and trial is uniquely seeded which allows for deterministic stochastic results between trials. When no seed is set, each algorithm and trial gets a random seed.

Users can either use the python `random` package to perform random operations or get framework specific generators using:
* `get_numpy_generator`
* `get_tensorflow_generator`
* `get_torch_generator`
* `get_next_jax_key`

all of which are seeded using the `iop.set_seed()` function. 

Two new random functions are added to the interoperability layer: `choice` and `rand`. Both work similarly to the numpy methods.

Minor change has been made to batch sampling for the default `EmpiricalRiskCost` implementation. Previously a uniform random subset of data was sampled, now a uniform random subset of data is sampled while removing those samples from subsequent batch samples until the entire dataset is exhausted. This better fits the expected behavior of batch sampling for machine learning applications. 

A bug was fixed where if you converted a PyTorch tensor into numpy and then back into PyTorch (performed in schemes) the dtype of the tensor got changed from `float32` to `float64`. This happened because the default dtype of numpy is `float64`. This was fixed by allowing to set the dtype of `to_[FRAMEWORK]` methods. `to_array_like` and similar methods were updated to extract the dtype from the `like` array. 

Other fixes and changes include: monolithic tests for the interoperability layer got split into multiple files, test for federated algorithms, update codebase to use RNG system, move citation to bibtex file, update warning messages and minor bugs.

Below is a short example of setting the seed for a benchmark run:
```python
import networkx as nx

import decent_bench.schemes as net_schemes
import decent_bench.utils.interoperability as iop
from decent_bench import benchmark
from decent_bench.agents import Agent
from decent_bench.costs import LogisticRegressionCost
from decent_bench.distributed_algorithms import (
    ADMM,
    ATC,
    DGD,
)
from decent_bench.networks import P2PNetwork

if __name__ == "__main__":
    iterations = 200
    trials = 2
    n_agents = 50

    iop.set_seed(123)

    costs, optimal_x, test_data = benchmark.create_classification_problem(
        LogisticRegressionCost,
        n_agents=n_agents,
    )
    graph = nx.complete_graph(n_agents)
    network = P2PNetwork(
        graph=graph,
        agents=[
            Agent(
                i,
                cost,
                activation=net_schemes.UniformActivationRate(0.5),
            )
            for i, cost in enumerate(costs)
        ],
        message_noise=net_schemes.GaussianNoise(0.0, 0.1),
        message_compression=net_schemes.Quantization(4),
        message_drop=net_schemes.UniformDropRate(0.01),
    )
    problem = benchmark.BenchmarkProblem(network, optimal_x, test_data)

    result = benchmark.benchmark(
        algorithms=[
            ADMM(iterations=iterations),
            DGD(iterations=iterations),
            ATC(iterations=iterations),
        ],
        benchmark_problem=problem,
        n_trials=trials,
    )
    met = benchmark.compute_metrics(result)
    benchmark.display_metrics(met)

```

closes #237

### DIFF
--- a/decent_bench/benchmark/_utils.py
+++ b/decent_bench/benchmark/_utils.py
@@ -187,9 +187,9 @@ def create_quadratic_problem(
     """
     A, b = [], []  # noqa: N806
     for _ in range(n_agents):
-        A_i = iop.uniform((size, size), framework=SupportedFrameworks.NUMPY, device=SupportedDevices.CPU)  # noqa: N806
+        A_i = iop.uniform(shape=(size, size), framework=SupportedFrameworks.NUMPY, device=SupportedDevices.CPU)  # noqa: N806
         A.append((A_i + iop.transpose(A_i)) / 2 + size * iop.eye_like(A_i))
-        b.append(iop.normal((size,), std=10, framework=SupportedFrameworks.NUMPY, device=SupportedDevices.CPU))
+        b.append(iop.normal(shape=(size,), std=10, framework=SupportedFrameworks.NUMPY, device=SupportedDevices.CPU))
 
     costs = [QuadraticCost(A[i], b[i]) for i in range(n_agents)]
     sum_cost = reduce(add, costs)
--- a/decent_bench/costs/_base/_regularizer_costs.py
+++ b/decent_bench/costs/_base/_regularizer_costs.py
@@ -220,7 +220,7 @@ class L1RegularizerCost(BaseRegularizerCost):
 
     def hessian(self, x: Array, **kwargs: Any) -> Array:  # noqa: ARG002, ANN401
         if self._hessian_cache is None:
-            self._hessian_cache = iop.zeros((self._dim, self._dim), framework=self.framework, device=self.device)
+            self._hessian_cache = iop.zeros(shape=(self._dim, self._dim), framework=self.framework, device=self.device)
         return self._hessian_cache
 
     def proximal(self, x: Array, rho: float, **kwargs: Any) -> Array:  # noqa: ARG002, ANN401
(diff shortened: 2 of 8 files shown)


decent_bench/datasets/_kaggle_handler.py
```diff
@@ -134,7 +132,7 @@ def get_partitions(self) -> Sequence[Dataset]:
 
     def _random_split(self, df: pd.DataFrame) -> Sequence[Dataset]:
         # Shuffle the dataframe
-        df = df.sample(frac=1, random_state=self.seed, replace=False).reset_index(drop=True)
+        df = df.sample(frac=1, random_state=random.randint(0, 2**32 - 1), replace=False).reset_index(drop=True)
```
nicola-bastianello: if I understand correctly, this would not be subject to the global seed even if it is set before the the creation of the dataset loader. but `PyTorchDatasetHandler` does use `get_torch_generator`, which is subject to the global seed if it is set beforehand

I took a quick look, and apparently `random_state` here can be set with a `np.random.Generator`. so here we could actually pass `iop.get_numpy_generator` and then also these random operations are subject to the global seed

PR #280, opened 2026-03-30
enh(display): refactor metrics filtering when displaying, and change plotting behavior for unavailable/diverged metrics

### LINKED ISSUE #215: Inconsistent tables and plots
Running the following made GT1 show up in the plots even though the table showed ``nan``. This should be looked into since nan values are expected to exclude the algorithm from the plots.

```
if __name__ == "__main__":
    problem = create_regression_problem(
        LogisticRegressionCost,
        n_agents=100,
        n_neighbors_per_agent=3,
        asynchrony=False,
        compression=False,
        noise=False,
        drops=True,
    )

    benchmark.benchmark(
        algorithms=[
            DGD(iterations=5000, step_size=0.001),
            GT1(iterations=5000, step_size=0.001),
            GT2(iterations=5000, step_size=0.001),
            ADMM(iterations=5000, rho=10, alpha=0.3),
        ],
        benchmark_problem=problem,
        n_trials = 30,
        confidence_level = 0.95,
        max_processes = 2,
        table_fmt="latex"
    )
```

### LINKED ISSUE #271: Customize which algorithms are shown in display_metrics
It would be nice to be able to customize which algorithms are shown in display_metrics similar to how you can change which metrics are shown. It should be simple, something like the following should work:

```python
def display_metrics(..., algorithms: type):
  ...
  
  if algorithms is not None:
    for alg in metrics_result.plot_results:
      if isinstance(alg, algorithms):
        new_metrics_result.plot_results.pop(alg, None)

  if table_metrics is not None:
    ...
  
```

Might need to change `Mapping` to `dict` in the `MetricResult` class for type checking.

nicola-bastianello:
This pull request introduces several improvements and robustness features to the metric computation and display pipeline in the benchmarking framework. The main changes include filtering out unavailable metrics, improved handling of divergent or incomplete data in plots, and new options for selectively displaying results for specific algorithms. These changes enhance the reliability of metric reporting and make the behavior of the system clearer when metrics cannot be computed.

**Motivation**

- Unavailable metrics (e.g. regret when x_optimal is not available) were silently returning nan; and were then displayed in tables with nan values, while being omitted from plots. This was confusing as the nan was an internal default, not something the user should be concerned with. Also, nan in tables should only signify divergence
- Available metrics computed on algorithms that diverge were omitted completely from plots, while they showed up in tables with nans. Nans in table are good, but they can also be plotted up to the point where they start getting nan/inf values (this gives a nice indication to the user that indeed the nan in the table is associated with divergence)

**Metric availability and filtering:**

* Added methods to the `Metric` class to mark a metric as unavailable during computation, including storing and exposing the reason, and clearing this status before new computations. This allows metrics to self-report when they cannot provide results (e.g., due to missing problem data).
* Updated `compute_metrics` to filter out metrics that have been marked unavailable, emitting warnings with omitted metric names. Plot and table metrics are now deep-copied to avoid state leakage.

**Plot and table computation robustness:**

* In `compute_plots` and `compute_tables`, metrics are checked for availability before computation. If a metric is unavailable or all trials diverge, it is omitted from results with a warning. Plot trajectories are truncated at the first non-finite or over-threshold point, and only the common finite prefix is used for aggregation.

**Display and selection improvements:**

* Added an `algorithms` parameter to `display_metrics`, allowing users to restrict displayed results to specific algorithms. The new `_filter_algorithms` helper filters all relevant result fields accordingly.

**Documentation and clarity:**

* Improved docstrings throughout the codebase to clarify the behavior of metric filtering, plot truncation, and requirements for certain metrics (e.g., `Regret` now documents that it requires `problem.x_optimal`).

closes #215 closes #271


Simpag:
The RTD failure is fixed in my latest PR. TLDR; something updated and now a warning message is printed because the class tagger extension is not thread safe, and warnings are treated as errors in our build config. I updated the build config to run using one process which solves it while keeping warning safety.

decent_bench/benchmark/_display.py
```diff
@@ -135,6 +143,36 @@ def display_metrics(
     )
 
 
+def _filter_algorithms(
+    metrics_result: MetricResult,
+    algorithms: list["Algorithm[Network]"],
+) -> MetricResult:
+    selected_names = {algorithm.name for algorithm in algorithms}
```
Simpag: Should mention that algorithms are filtered by name. Since we allow for multiple of the same algorithm to be benchmarked, they should name them uniquely.

decent_bench/benchmark/_display.py
```diff
@@ -135,6 +143,36 @@ def display_metrics(
...
+
+    if metrics_result.agent_metrics is not None:
+        metrics_result.agent_metrics = {
+            algorithm: metrics
+            for algorithm, metrics in metrics_result.agent_metrics.items()
+            if algorithm in algorithms or algorithm.name in selected_names
```
Simpag: I dont think `algorithm in algorithms` will ever be true since it is most likely a new algorithm object. It can be the same in rare cases when users reuse the same list of algorithm during the same python execution. But I cant see why a user would benchmark X algorithms but only want to plot Y algorithms. 

This applies to checks below too. You can decide if you want to keep it or not, it is not a huge computational cost

decent_bench/benchmark/_compute.py
```diff
@@ -19,6 +17,9 @@
     from decent_bench.utils.checkpoint_manager import CheckpointManager
 
 
+PlotMetricData = tuple[Sequence[float], Sequence[float], Sequence[float], Sequence[float]]
```
Simpag: I would either move this to `utils.types` or remove it. If you decide to keep it then you can also update `MetricResult` class

decent_bench/benchmark/_compute.py
```diff
@@ -110,3 +124,49 @@ def compute_metrics(
...
+    table_results: Mapping[Algorithm[Network], Mapping[Metric, Mapping[str, tuple[float, float]]]],
+) -> list[Metric]:
+    filtered_metrics = [
+        metric for metric in table_metrics if any(metric in results for results in table_results.values())
+    ]
+    omitted = [metric for metric in table_metrics if metric not in filtered_metrics]
+    if omitted:
+        logger.LOGGER.warning(
+            f"Omitting unavailable table metrics: {', '.join(metric.table_description for metric in omitted)}"
+        )
```
Simpag: Isnt this warning message already printed in `compute_tables`?

decent_bench/benchmark/_compute.py
```diff
@@ -110,3 +124,49 @@ def compute_metrics(
...
+        omitted: list[Metric] = []
+        for group in grouped_plot_metrics:
+            filtered_group = [metric for metric in group if metric in available_metrics]
+            omitted.extend(metric for metric in group if metric not in available_metrics)
+            if filtered_group:
+                filtered_groups.append(filtered_group)
+        if omitted:
+            logger.LOGGER.warning(
+                f"Omitting unavailable plot metrics: {', '.join(metric.plot_description for metric in omitted)}"
+            )
```
Simpag: Isnt this warning alreay printed in `compute_plots`?

decent_bench/benchmark/_compute.py
```diff
@@ -110,3 +124,49 @@ def compute_metrics(
...
+        return filtered_groups
+
+    flat_plot_metrics = cast("list[Metric]", plot_metrics)
+    filtered_metrics = [metric for metric in flat_plot_metrics if metric in available_metrics]
+    omitted = [metric for metric in flat_plot_metrics if metric not in available_metrics]
+    if omitted:
+        logger.LOGGER.warning(
+            f"Omitting unavailable plot metrics: {', '.join(metric.plot_description for metric in omitted)}"
+        )
```
Simpag: Isnt this already printed in `compute_plots`?

Simpag:
> one thing to discuss: currently, the `metric_utils` exposes several utilities in the public API (regret, accuracy, etc). since they are exposed, they check conditions and return nan when not met. however, the same is now done in `Metric` (which marks itself unavailable when conditions are not met, and returns nan). this leads to redundant checks
> 
> I see two options:
> 
>     1. the checks are abstracted to some private utils, to reduce code duplication
> 
>     2. we remove from the public API the utilities used to compute the metrics
> 
> 
> I personally think 2. would be the best solution; the utilities are not really something users will need to have access to (as we already provide the respective metrics). also, the way these utilities are coded is also not very user friendly, so they have not a lot of educational importance

I would update the utility functions so instead of returning nan it raises an error then do the "mark_unavailable" based on which error was raised. I would like to keep these metric utils available to users as there might be some cases where a user wants to create some custom metric by combining different metrics in some way

The final metrics would be something like:
```python
try:
    return utils.accuracy(agents, problem, iteration)
except utils.MetricUnavailableException as e:
    self.mark_unavailable(str(e))
    return [float("nan") for _ in agents]
```
which would require a custom exception to be created, something like this should be sufficient:
```python
class MetricUnavailableException (Exception):
    """docstring"""

    def __init__(self, message):
        super().__init__(message)

    def __str__(self):
        return f"{self.message}"
```

EDIT: It is not required to create a custom exception but it keeps the code a little cleaner and wont accidentally mask any random errors

decent_bench/metrics/_plots.py
```diff
@@ -171,28 +188,57 @@ def compute_plots(
         )
 
         for metric in flat_metrics:
+            metric.clear_unavailable()
             progress.update(plot_task, status=f"Task: {metric.plot_description}")
```
Simpag: I would probably add the metric availability check to before progress.update() so that we dont waste a bunch of compute. This can be dome by doing:
```python
metric.clear_unavailable()
metric.get_data_from_trial(resulting_agent_states[algs[0]][0], problem, iteration=-1)
if metric.is_unavailable:
    LOGGER.warning("ERROR MSG")
    progress.advance(plot_task, advance=len(resulting_agent_states))
    continue
progress.update(plot_task, status=f"Task: {metric.plot_description}")
```

I dont think the availability checks depend on the iteration or algorithm, just the benchmark problem.

decent_bench/metrics/_plots.py
```diff
@@ -171,28 +188,57 @@ def compute_plots(
...
+                    reason = metric.unavailable_reason or "unknown reason"
+                    LOGGER.warning(
+                        f"Skipping plot metric '{metric.plot_description}' because it is unavailable: {reason}"
                     )
-                    LOGGER.warning(msg)
+                    for result in results.values():
+                        result.pop(metric, None)
+                    remaining = len(resulting_agent_states) - alg_idx
+                    for _ in range(remaining):
+                        progress.advance(plot_task)
```
Simpag: Use the parameter `advance` in `progress.advance` instead of a for loop. (But might remove this if you decide to do the earlier checking)

decent_bench/metrics/_tables.py
```diff
@@ -143,9 +143,19 @@ def compute_tables(
...
             progress.update(table_task, status=f"Task: {metric.table_description}")
             data_per_trial = [_table_data_per_trial(resulting_agent_states[a], problem, metric) for a in algs]
 
+            if metric.is_unavailable:
+                reason = metric.unavailable_reason or "unknown reason"
+                LOGGER.warning(
+                    f"Skipping table metric '{metric.table_description}' because it is unavailable: {reason}"
+                )
+                for _ in metric.statistics:
+                    progress.advance(table_task)
```
Simpag: Use the parameter `advance` in `progress.advance` instead of a for loop. (But might remove this if you decide to do the earlier checking)

decent_bench/metrics/_tables.py
```diff
@@ -143,9 +143,19 @@ def compute_tables(
             status="",
         )
         for metric in metrics:
+            metric.clear_unavailable()
```
Simpag: I would also update the availability check here so we dont waste compute, while this is much cheaper than plot metrics its still a nice feature.
```python
metric.clear_unavailable()
metric.get_data_from_trial(resulting_agent_states[algs[0]][0], problem, iteration=-1)
if metric.is_unavailable:
    LOGGER.warning("ERROR MSG")
    progress.advance(table_task, advance=len(metric.statistics))
    continue
progress.update(table_task, status=f"Task: {metric.table_description}")
```

Simpag (commented):
Overall nice. Just some minor refactoring, performance improvements and possibly duplicated warning messages.

nicola-bastianello:
> I would update the utility functions so instead of returning nan it raises an error then do the "mark_unavailable" based on which error was raised. I would like to keep these metric utils available to users as there might be some cases where a user wants to create some custom metric by combining different metrics in some way

yes, great point, I'll go for this option. and I also like having a custom error

decent_bench/benchmark/_display.py
nicola-bastianello: yes, I forgot to mention this was another thing to discuss. I see a few options to handle this:

1. we check by name like now, and document this behavior; I think the cleanest way to do this would be to check at init of `benchmark`, so that we don't raise after a lot of compute has been wasted
2. if two algorithms have the same name, we append `(1)`, `(2)`, etc to it. I don't much like this as it could be surprising to users
3. we implement `__eq__` for `Algorithm` (we set `@dataclass(eq=True)`), and two algorithms become the same if all their attributes are the same, including the name. then we can check via `==` and everything works out as long as all the algorithms are different

I'm leaning towards 1. as it is quite a quick, it is reasonable, and can help avoid unintuitive behaviors (e.g. seeing the same algorithm name twice in a legend, due to the user's mistaken copy-paste for example). 2. as I said I don't like, and 3. is more computationally heavy and could still lead to having the same algorithm name twice in legend

decent_bench/benchmark/_display.py
nicola-bastianello: I agree, I'll remove it

decent_bench/benchmark/_compute.py
nicola-bastianello: I think I added it when debugging mypy errors, I think I can safely remove it

decent_bench/benchmark/_compute.py
nicola-bastianello: I'll double check there is no repetition of warnings

decent_bench/metrics/_plots.py
nicola-bastianello: good point, thank you, I'll do this

decent_bench/benchmark/_display.py
Simpag: I think option 1 is the best. Just raise an error if two algorithms have the same name

decent_bench/metrics/_plots.py
nicola-bastianello: I'm not sure this works, because `algs[0]][0]` assumes that at least one algorithm is present, and that that algorithm has at least one trial. but if I understand correctly this might not be the case

I'll look into it some more in the next few days

Simpag: If no algorithms are present then there isnt anything to plot. This for loop would not be started so that is safe. Not having any trials should not be possible as you need at least one trial to run a benchmark, the program should have crashed somewhere else at that point.

Simpag:
Could maybe allow users to pass a list of strings to `display_metrics` too, users might have a separate script to look at the plots, should be a very minor change. Just add a check before display the metrics to validate that at least one algorithm and one metric is present

nicola-bastianello:
> Could maybe allow users to pass a list of strings to `display_metrics` too, users might have a separate script to look at the plots, should be a very minor change. Just add a check before display the metrics to validate that at least one algorithm and one metric is present

good idea; we can allow passing a list of algorithm's names, now that they are guaranteed to be unique. however, I'm not sure about metrics: I could use `MetricSubclass.__name__`, but there might be more than one (e.g. same metric but with different statistics), so it could get ambiguous

adrianardv:
The failing mypy check seems to come from pre-existing issues on main. I ran into the same issue on my open PR and fixed those mypy errors there.

nicola-bastianello:
> The failing mypy check seems to come from pre-existing issues on main. I ran into the same issue on my open PR and fixed those mypy errors there.

thanks! then I'll merge your PR before I merge this one

decent_bench/metrics/_plots.py
nicola-bastianello: thanks for clarifying

thinking some more about it, I'm afraid that the availability check based on utils errors might not be the most flexible option. indeed, now we need to do `get_data_from_trial` which sets the availability status if the respective util raises. I think it would make more sense to have `Metric.is_available(benchmark_problem)`, which can check availability very easily and without having to call the utils

this would require privating the utils. I've thought about it, and I don't see why exposing them in the API would be helpful: 1) if an advanced user wants to access the utils they can still inspect the code; 2) I don't see how these utils would be combined into more complex metrics; 3) I find it a bit confusing to expose the Metric and also its util. only exposing the metric would make it clear that that's what users should care about

Simpag: So if I understand you correctly, you want to move the availability check to a new method in the metric so that it can be called before we even perform compute tables/plots? While this is cleaner and can be more performant it does make it a bit harder to implement new metrics. I don’t mind either solution

Simpag:
> 
> good idea; we can allow passing a list of algorithm's names, now that they are guaranteed to be unique. however, I'm not sure about metrics: I could use `MetricSubclass.__name__`, but there might be more than one (e.g. same metric but with different statistics), so it could get ambiguous

I believe that table and plot metrics are compared by table and plot descriptions respectively. Statistics cannot be changed as they are currently computed in `compute_metrics`. Its not perfect but I would assume that users almost always just creates one metric of each class. Might want to create a check for it as it makes a lot more sense to add more statistics to the same object instead of having multiple instances, its also much more efficient.

decent_bench/metrics/_plots.py
nicola-bastianello: yes, that's what I have in mind. in my view, only advanced users will implement new metrics, so I think an additional method for availability is not a big problem. I also think it will be a good think to force users to implement this, as it makes them aware of what attributes BenchmarkProblem needs to contain

Simpag: Sounds good. It simplifies it a bit, with this you’d be able to remove the metrics before computation and wont have to do anything advanced to account fir the progress bar.

nicola-bastianello:
the legend placement is now conditional on the number of labels:

- below a certain number, the legend is drawn in the same figure as the plots (at the top, outside of the plots -- same as before)
- above that number, the legend is drawn in a separate figure (which is stored with `-legend` appended to the file name). the size of the legend is probed by drawing it, so that the figure is accurately sized to fit it

plot
plot_legend

decent_bench/benchmark/_benchmark.py
```diff
@@ -121,6 +130,7 @@ def resume_benchmark(  # noqa: PLR0912
             algorithms = checkpoint_manager.load_initial_algorithms()
             if algorithms is None:
                 raise ValueError("Initial algorithms not found in checkpoint metadata")
+            _validate_unique_algorithm_names(algorithms)
```
Simpag: It doesnt matter but this should always pass. Users cant change the algorithms from a checkpoint

decent_bench/utils/_metric_helpers.py
```diff
@@ -0,0 +1,23 @@
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from decent_bench.metrics import Metric
+
+
+def _flatten_plot_metrics(plot_metrics: list["Metric"] | list[list["Metric"]]) -> list["Metric"]:
+    if any(isinstance(metric, list) for metric in plot_metrics):
```
Simpag, line 8: Should add a check that if any are lists, they all must be lists

decent_bench/benchmark/_display.py
```diff
@@ -117,41 +131,90 @@ def display_metrics(
...
-        compare_iterations_and_computational_cost=compare_iterations_and_computational_cost,
-        individual_plots=individual_plots,
-        plot_grid=plot_grid,
-        plot_format=plot_format,
-        plot_path=save_path,
-    )
+    if new_metrics_result.table_metrics:
+        display_tables(new_metrics_result, table_fmt=table_fmt, scale_compute=scale_compute, table_path=save_path)
+    else:
+        LOGGER.warning("No table metrics to display, skipping tables")
```
Simpag, line 161: These warnings are already check in `display_X`. Consider removing the warning here or in `display_X`

decent_bench/benchmark/_metric_result.py
```diff
@@ -41,3 +43,53 @@ class MetricResult:
         ]
         | None
     )
+
+    @property
+    def available_algorithms(self) -> list[str]:
+        """Return ``name`` of available algorithms, which can be used for filtering in :func:`~decent_bench.benchmark.display_metrics`."""  # noqa: E501
+        names: list[str] = []
+        seen: set[str] = set()
```
Simpag: You dont need to store a list. You could just return `sorted(seen)`. You could also make this a cached property but its up to you, Im not sure if it will be called more than once and its very fast

decent_bench/benchmark/_metric_result.py
```diff
@@ -41,3 +43,53 @@ class MetricResult:
...
+            return []
+
+        descriptions: list[str] = []
+        seen: set[str] = set()
+        for metric in self.table_metrics:
+            if metric.table_description not in seen:
+                seen.add(metric.table_description)
+                descriptions.append(metric.table_description)
+
+        return descriptions
```
Simpag: Can replace with `return sorted({metric.table_description for metric in self.table_metrics})`

decent_bench/metrics/_metric.py
```diff
@@ -82,6 +82,28 @@ def can_diverge(self) -> bool:
         """
         return True
 
+    def is_available(
```
Simpag, line 85: Might want to make this an abstract method so that users always has to think about this?

decent_bench/metrics/_plots.py
```diff
@@ -172,6 +190,12 @@ def compute_plots(
 
         for metric in flat_metrics:
             progress.update(plot_task, status=f"Task: {metric.plot_description}")
+            available, reason = metric.is_available(problem)
+
+            if not available:
+                LOGGER.warning(f"Skipping plot metric '{metric.plot_description}' because it is unavailable: {reason}")
+                progress.advance(plot_task, advance=len(resulting_agent_states))
+                continue
```
Simpag: While the availability check is usually inexpensive, we might want to move them to `compute_metrics` method instead so we dont check them twice. This also makes the progress bars smoother as it wont skip when incompatible metrics are passed

Simpag (commented):
Overall looks good, just some minor code simplifications

decent_bench/metrics/metric_utils.py
```diff
@@ -30,7 +29,10 @@ class MetricProgressBar(Progress):
 
     def __init__(self) -> None:
         super().__init__(
-            TextColumn("[progress.description]{task.description}"),
+            TextColumn(
+                "[progress.description]{task.description}",
+                table_column=Column(width=24, no_wrap=True),
```
Simpag, line 34: Maybe set max_width instead of width? Or did you hard code the width so that the progress bars are aligned?

Simpag:
I would request I re-review from co-pilot to check if it can find any bugs once you feel like you're done. A lot has changed since the first review and its really good at spotting small mistakes even if it says a lot of garbage.

decent_bench/benchmark/_display.py
nicola-bastianello: good catch, I'll remove from `display_X`, as I think it's clearer to not even launch the `display_X` functions if there's nothing to plot

decent_bench/metrics/_metric.py
nicola-bastianello: I thought about it but ultimately decided against it, so that we can rely on the super method for Metrics that are always available

decent_bench/metrics/metric_utils.py
nicola-bastianello: indeed, I hardcoded to align them. do you see some negative effect of this?

decent_bench/benchmark/_metric_result.py
nicola-bastianello: I thought about caching, but I also thnk it will be called only once so didn't see the point. copilot was also warning against potential issues coming from caching (I forget exactly what)

decent_bench/metrics/_plots.py
nicola-bastianello: yes, good point! I'll refactor so that both table and plot metrics are checked `compute_metrics`. I also realized the plot_metrics are flattened quite a few times, I'll try to reduce that number or reuse the results of flattening

decent_bench/metrics/_metric.py
Simpag: Sounds good. I guess we’ll have docs for how to create each module of the pipeline so this will be mentioned there

decent_bench/metrics/metric_utils.py
Simpag: No, I think it’s fine. Only thing that might happen is that the progress bar can get very short on narrow terminal windows but that’s uncommon

decent_bench/metrics/_plots.py
```diff
@@ -180,19 +197,35 @@ def compute_plots(
                     metric,
                 )
```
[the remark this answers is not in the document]
nicola-bastianello: unavailable metrics are filtered out in `compute_metrics`, which then calls `compute_plots`. so no need to guard in the latter

decent_bench/metrics/metric_utils.py
```diff
@@ -104,35 +104,24 @@ def gradient_norm(agents: Sequence[AgentMetricsView], iteration: int = -1) -> fl
     return float(la.norm(grad_avg)) ** 2
```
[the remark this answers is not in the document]
nicola-bastianello: this refactor is on purpose, the entry point are always `Metric` objects, so decided not to expose utils

decent_bench/metrics/metric_utils.py
```diff
@@ -104,35 +104,24 @@ def gradient_norm(agents: Sequence[AgentMetricsView], iteration: int = -1) -> fl
...
-    x_per_iteration = np.asarray([
-        iop.to_numpy(x) for iteration, x in sorted(agent.x_history.items()) if iteration <= up_to_iteration
-    ])
+    """
+    agent_iteration = agent.x_history.max() if iteration == -1 else iteration
+    x_at_iteration = iop.to_numpy(agent.x_history[agent_iteration])
     opt_x = iop.to_numpy(problem.x_optimal)  # type: ignore[arg-type]
-    errors: NDArray[float64] = la.norm(x_per_iteration - opt_x, axis=tuple(range(1, x_per_iteration.ndim)))
-    return errors
+    return float(la.norm(x_at_iteration - opt_x))
```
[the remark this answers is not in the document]
nicola-bastianello: this is on purpose to align `_x_error` to the other utils, and `XError` to the other metrics

decent_bench/metrics/_tables.py
```diff
@@ -144,6 +143,7 @@ def compute_tables(
         )
         for metric in metrics:
             progress.update(table_task, status=f"Task: {metric.table_description}")
+
```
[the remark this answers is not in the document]
nicola-bastianello: no, same as `compute_plots`, the check happens in `compute_metrics`, which is the only thing that will directly be called by users

decent_bench/metrics/_plots.py
```diff
@@ -44,6 +45,14 @@
...
 MARKERS = ["o", "s", "v", "^", "*", "D", "H", "<", ">", "p", "P", "X"]
 STYLES = ["-", ":", "--", "-.", (5, (10, 3)), (0, (5, 10)), (0, (3, 1, 1, 1))]
+# Upper bound used when deciding whether a point is still plottable on log y-axis.
+# Values above this threshold are treated as divergence for plotting and trigger truncation.
+MAX_LOG_PLOT_VALUE = 1e100
+# thresholds to decide when to plot the legend as a separate figure
+LEGEND_MAX_COLS = 3
+SEPARATE_LEGEND_MIN_LABELS = 9
+SEPARATE_LEGEND_MIN_ROWS = 4
+SEPARATE_LEGEND_MAX_LABEL_LENGTH = 24
```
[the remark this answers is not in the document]
nicola-bastianello: changed name and comment; it's better to apply also to linear scale, as inf can create issues in linear scale as well

decent_bench/benchmark/_display.py
```diff
@@ -160,44 +223,46 @@ def _get_new_table_metrics(
...
+            original_metric = plot_lookup.get(metric_name)
             if original_metric is None:
-                LOGGER.warning(
-                    f"Requested plot metric '{group.plot_description}' not found in metrics result, skipping"
-                )
+                LOGGER.warning(f"Requested plot metric '{metric_name}' not found in metrics result, skipping")
                 continue
             new_plot_metrics.append(original_metric)  # type: ignore[arg-type]
 
     return new_plot_metrics
```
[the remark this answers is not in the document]
nicola-bastianello: added the check

decent_bench/benchmark/_benchmark.py
```diff
@@ -101,6 +109,7 @@ def resume_benchmark(  # noqa: PLR0912
     Raises:
         ValueError: If the checkpoint directory does not exist, is empty, or contains invalid metadata.
         ValueError: If increase_iterations or increase_trials is negative.
+        ValueError: If any two algorithms loaded from the checkpoint share the same name.
```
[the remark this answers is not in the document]
nicola-bastianello: changed docstring

decent_bench/benchmark/_benchmark.py
```diff
@@ -35,6 +35,14 @@
     from decent_bench.utils.progress_bar import ProgressBarHandle
 
 
+def _validate_unique_algorithm_names(algorithms: list[Algorithm[Network]]) -> None:
+    names = [alg.name for alg in algorithms]
+    duplicate_names = sorted({n for n in names if names.count(n) > 1})
+    if duplicate_names:
+        duplicates = ", ".join(duplicate_names)
+        raise ValueError(f"Algorithm names must be unique, duplicates found: {duplicates}")
```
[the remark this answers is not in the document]
nicola-bastianello: done

docs/source/user.rst
```diff
@@ -189,6 +189,45 @@ If you set ``plot_metrics`` to a list of lists of :class:`~decent_bench.metrics.
...
+~~~~~~~~~~~~~~~~~~~
+Some metrics require additional problem information.
+
+- ``regret`` and ``x error`` require ``problem.x_optimal``.
+- ``accuracy``, ``mse``, ``precision``, and ``recall`` require ``problem.test_data`` and
+    agents with :class:`~decent_bench.costs.EmpiricalRiskCost`.
+- ``accuracy``, ``precision``, and ``recall`` additionally require integer-valued targets.
+
+If these requirements are not met, the metric marks itself unavailable, returns NaN, and is omitted from the final
+metric lists returned by :func:`~decent_bench.benchmark.compute_metrics`.
```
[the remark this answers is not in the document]
nicola-bastianello: done

decent_bench/benchmark/_compute.py
```diff
@@ -62,8 +63,14 @@ def compute_metrics(
         All used table- and plot-metrics will be saved to the checkpoints' metadata if a checkpoint manager is provided,
         in order to know which metrics were computed and can be displayed later.
 
+        Metrics that return ``False`` from :meth:`~decent_bench.metrics.Metric.is_available` for the given problem
+        filtered out from the returned metric lists. Warnings are emitted with the omitted metric names.
```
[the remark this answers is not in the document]
nicola-bastianello: done

decent_bench/metrics/metric_library.py
```diff
@@ -394,19 +409,39 @@ class Accuracy(Metric):
 
     where TP, TN, FP, and FN are true positives, true negatives, false positives, and false negatives, respectively.
 
+    Note:
+        Available only when:
+
+        - ``problem.test_data`` is provided,
+        - all agent costs are :class:`~decent_bench.costs.EmpiricalRiskCost`,
+        - target labels are integer-valued.
```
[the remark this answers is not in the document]
nicola-bastianello: done

decent_bench/metrics/metric_library.py
Simpag, line 166: Might be good to mention which norm is used, I think iop.norm defaults to L2 norm

Simpag (commented):
LGTM, just a minor addition to a doc string

PR #281, opened 2026-03-30
fix(Multiprocessing): Fully disable multiprocessing when not used

### LINKED ISSUE #279: Windows-specific pytest slowdown in `test/utils/test_checkpoints.py`
Running the test suite locally has become very slow. It seems to be due to the changes introduced in the commit `feat(RNG): Cross-framework rng (#274)`.

I've checked and the slowdown seems to be concentrated in: `pytest test\utils\test_checkpoints.py -q`. 
That file takes several minutes on my machine and feels slow throughout. 

Is this happening on your machines too? Is this expected, or should we try to reduce runtime in `test/utils/test_checkpoints.py`?

Simpag:
This PR fully disables multiprocessing when `max_processes = 1`. This speeds up execution on windows based systems as the overhead of spawning processing using `spawn` context is large. For this reason, this PR also disabled the multiprocessing checkpointing tests for all systems not running Linux since Linux by default uses `fork` which comes with less overhead. Finally, a better error message is given when users runs unguarded benchmarks (main script is not wrapped in `if __name__ == '__main__'`) while using multiprocessing (`max_processes > 1`).  

closes #279

### DIFF
--- a/decent_bench/benchmark/_benchmark.py
+++ b/decent_bench/benchmark/_benchmark.py
@@ -268,9 +268,6 @@ def benchmark(
             "Progress cannot be resumed if interrupted."
         )
 
-    if runtime_metrics is not None and len(runtime_metrics) == 0:
-        runtime_metrics = None
-
     results = _benchmark(
         algorithms=algorithms,
         benchmark_problem=benchmark_problem,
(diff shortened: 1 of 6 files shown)


adrianardv:
Maybe it would also be worth adding a short note in the user guide about the Windows spawn behavior and the if __name__ == "__main__": guard?

Simpag:
> Maybe it would also be worth adding a short note in the user guide about the Windows spawn behavior and the if **name** == "**main**": guard?

Good idea, the user docs are *very* outdated but its still good so we remember to include it

Simpag:
I have not touched anything with the documentation and RTD fails... Why is complaining now about the class tagger?

nicola-bastianello:
thank you for fixing this and the RTD issue!

I confirm the tests now take very little time on my machine too (<10s). merging now

Issue #273, opened 2026-03-27
Refactor interoperability


Simpag:
Currently the interoperability layer is a bit dense and hard to extend. My suggestion is to turn the interoperability layer into a class based system. I would define an abstract class containing all the methods which we currently support in the interoperability layer. Then each framework can implement all of these operations and register to the main handler. Something like the following:

```python
from abc import ABC, abstractmethod
from typing import TYPE_CHECKING, Any

import numpy as np

from decent_bench.utils.array import Array
from decent_bench.utils.types import SupportedArrayTypes, SupportedFrameworks

if TYPE_CHECKING:
    import torch

try:
    import torch

    TORCH_AVAILABLE = True
except ImportError:
    TORCH_AVAILABLE = False


class Interop(ABC):
    @abstractmethod
    def sum(self, array: Any, dim: int | tuple[int, ...] | None = None, keepdims: bool = False) -> Any:
        """Sum of array elements over a given axis."""


class NumpyIterop(Interop):
    def sum(
        self,
        array: np.ndarray[np.float64],
        dim: int | tuple[int, ...] | None = None,
        keepdims: bool = False,
    ) -> np.ndarray[np.float64]:
        res: np.ndarray[np.float64] = np.sum(array, axis=dim, keepdims=keepdims)
        return res


class TorchInterop(Interop):
    def sum(
        self,
        array: torch.Tensor,
        dim: int | tuple[int, ...] | None = None,
        keepdims: bool = False,
    ) -> torch.Tensor:  # Replace with actual torch.Tensor type when available
        return torch.sum(array, dim=dim, keepdim=keepdims)


class Interoperability(Interop):
    def __init__(self) -> None:
        super().__init__()
        self.backends: dict[SupportedFrameworks, Interop] = {}
        self.types: dict[SupportedFrameworks, tuple[type, ...]] = {}

    def register_backend(
        self, framework: SupportedFrameworks, backend: Interop, backend_types: tuple[type, ...]
    ) -> None:
        self.backends[framework] = backend
        self.types[framework] = backend_types

    def framework_of_array(self, array: SupportedArrayTypes) -> SupportedFrameworks:
        value = array.value if isinstance(array, Array) else array

        for framework, types in self.types.items():
            if isinstance(value, types):
                return framework

        raise TypeError(f"Unsupported framework type: {type(value)}")

    def _return_array(self, array: SupportedArrayTypes) -> Array:
        if not TYPE_CHECKING:
            return array

        return Array(array)

[...]

PR #282, opened 2026-04-01
fix: enforce backend consistency and expand tests

### LINKED ISSUE #275: Explicitly allow or disallow agents with costs in different frameworks
At this time we don't explicitly allow or disallaw different agents to have costs in different frameworks:

- In some places we raise when the frameworks are mismatched (like when using `Cost._validate_cost_operation(self, other, check_framework: bool = True)`). This would probably disallow mismatched frameworks when e.g. computing the sum of the local costs.
- In other places, we implicitly assume that all costs have the same framework, like when we set `SumCost.framework` as `return self.costs[0].framework`. This doesn't explicitly disallow mismatched frameworks, but silently ignores it

Overall, I think we should decide for a clear behavior and implement it everywhere: either we allow or we disallow mismatched frameworks, and ensure that the docs clarify this.

Thanks to the iop layer and the Array object, I think it's not impossible to allow mismatched frameworks, but there are some places like `SumCost` where it's not clear what we should do.

adrianardv:
### Summary

This PR adds a set of small enhancements and test coverage improvements across costs, networks, and federated behavior.

The main functional change is to enforce backend consistency more strictly:
- disallow networks whose agents use costs with different shape, framework, or device
- disallow composing costs with different framework or device

This addresses the main issue in #275. The only remaining point discussed there is globally setting framework/device through IOP.

### What changed

- enforce cost compatibility checks in network construction
- enforce framework/device compatibility in generic cost composition and `SumCost`
- clarify FedAvg local update routing so empirical-risk-preserving costs keep minibatch behavior while generic wrapped costs use full gradients
- document `PyTorchCost` + built-in regularizer
- extend regression coverage for costs, cost operators, networks, and federated behavior

----

Closes #275

### DIFF
--- a/decent_bench/distributed_algorithms.py
+++ b/decent_bench/distributed_algorithms.py
@@ -179,6 +179,10 @@ class FedAlgorithm(Algorithm[FedNetwork]):
     def _cleanup_agents(self, network: FedNetwork) -> Iterable["Agent"]:
         return [network.server(), *network.clients()]
 
+    def _sync_server_to_clients(self, network: FedNetwork, selected_clients: Sequence["Agent"]) -> None:
+        """Send the current server model to the selected clients."""
+        network.send(sender=network.server(), receiver=selected_clients, msg=network.server().x)
+
     def select_clients(
         self,
         clients: Sequence["Agent"],
@@ -327,16 +331,13 @@ class FedAvg(FedAlgorithm):
         self._run_local_updates(network, selected_clients)
         self.aggregate(network, selected_clients)
 
-    def _sync_server_to_clients(self, network: FedNetwork, selected_clients: Sequence["Agent"]) -> None:
-        network.send(sender=network.server(), receiver=selected_clients, msg=network.server().x)
-
     def _run_local_updates(self, network: FedNetwork, selected_clients: Sequence["Agent"]) -> None:
         for client in selected_clients:
             client.x = self._compute_local_update(client, network.server())
             network.send(sender=client, receiver=network.server(), msg=client.x)
 
     @staticmethod
-    def _cost_for_minibatch_local_updates(cost: Cost) -> EmpiricalRiskCost | None:
+    def _empirical_cost_for_local_updates(cost: Cost) -> EmpiricalRiskCost | None:
         """
         Return the empirical cost view to use for FedAvg local mini-batching.
 
@@ -351,32 +352,27 @@ class FedAvg(FedAlgorithm):
 
     def _compute_local_update(self, client: "Agent", server: "Agent") -> "Array":
         local_x = iop.copy(client.messages[server]) if server in client.messages else iop.copy(client.x)
-        empirical_cost = self._cost_for_minibatch_local_updates(client.cost)
+        empirical_cost = self._empirical_cost_for_local_updates(client.cost)
         if empirical_cost is not None:
-            return self._epoch_minibatch_update(
-                empirical_cost,
-                local_x,
-                empirical_cost.batch_size,
-                empirical_cost.n_samples,
-            )
+            return self._run_minibatch_local_epochs(empirical_cost, local_x)
+        return self._run_full_gradient_local_epochs(client.cost, local_x)
 
+    def _run_full_gradient_local_epochs(self, cost: Cost, local_x: "Array") -> "Array":
         for _ in range(self.num_local_epochs):
-            grad = client.cost.gradient(local_x)
+            grad = cost.gradient(local_x)
             local_x -= self.step_size * grad
         return local_x
 
-    def _epoch_minibatch_update(
+    def _run_minibatch_local_epochs(
         self,
         cost: EmpiricalRiskCost,
         local_x: "Array",
-        per_client_batch: int,
-        n_samples: int,
     ) -> "Array":
         for _ in range(self.num_local_epochs):
-            indices = list(range(n_samples))
+            indices = list(range(cost.n_samples))
             random.shuffle(indices)
-            for start in range(0, n_samples, per_client_batch):
-                batch_indices = indices[start : start + per_client_batch]
+            for start in range(0, cost.n_samples, cost.batch_size):
+                batch_indices = indices[start : start + cost.batch_size]
                 grad = cost.gradient(local_x, indices=batch_indices)
                 local_x -= self.step_size * grad
         return local_x
--- a/decent_bench/distributed_algorithms.py
+++ b/decent_bench/distributed_algorithms.py
@@ -6,7 +6,7 @@ from typing import TYPE_CHECKING, Any, Final, cast, final
 
 import decent_bench.utils.algorithm_helpers as alg_helpers
 import decent_bench.utils.interoperability as iop
-from decent_bench.costs import EmpiricalRiskCost
+from decent_bench.costs import Cost, EmpiricalRiskCost
 from decent_bench.networks import FedNetwork, Network, P2PNetwork
 from decent_bench.schemes import ClientSelectionScheme, UniformClientSelection
 from decent_bench.utils._tags import tags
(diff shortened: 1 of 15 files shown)


nicola-bastianello:
looks good, thank you! I'll review the details soon

one idea I had: currently the checks are made using Cost's method `_validate_cost_operation`, which only does a pairwise check between self and another cost. but now that we have decided on enforcing consistency, I think this utility could be spun off into a iop utils. it could be called `verify_compatible_backends` or something like that, and it takes a list of costs to check (when #273 is addressed, this util could then be modified to allow for a single cost also, which is checked against the backend set by `set_backend`). this way we have a single clear place where we document this backend consistency and easy to maintain

I would say you can set this PR to close #275 , and we address the `set_backend` with #273

decent_bench/distributed_algorithms.py
```diff
@@ -179,6 +179,10 @@ class FedAlgorithm(Algorithm[FedNetwork]):
     def _cleanup_agents(self, network: FedNetwork) -> Iterable["Agent"]:
         return [network.server(), *network.clients()]
 
+    def _sync_server_to_clients(self, network: FedNetwork, selected_clients: Sequence["Agent"]) -> None:
```
nicola-bastianello: if we want to make this a method in `FedAlgorithm` maybe it should be public, so that users implementing new algorithms can benefit from this util. if we want to keep private, then it should probably be in the subclass

if made public, could be renamed like `server_broadcast` or something like that (for shorter name)

nicola-bastianello (commented):
looks good, thanks! just one comment

decent_bench/distributed_algorithms.py
```diff
@@ -327,38 +331,48 @@ def step(self, network: FedNetwork, iteration: int) -> None:  # noqa: D102
...
-            indices = list(range(n_samples))
+            indices = list(range(cost.n_samples))
             random.shuffle(indices)
-            for start in range(0, n_samples, per_client_batch):
-                batch_indices = indices[start : start + per_client_batch]
+            for start in range(0, cost.n_samples, cost.batch_size):
+                batch_indices = indices[start : start + cost.batch_size]
                 grad = cost.gradient(local_x, indices=batch_indices)
                 local_x -= self.step_size * grad
         return local_x
```
Simpag, line 348: While this assures that a full epoch through the dataset is performed, I've updated empirical risk cost sampling so that it will iterate through the entire dataset (random order) before it re-uses datapoints. Therefore you could simplify this a bit by removing the indices parameter, but its not a big deal. 

One side effect; if you're using PyTorchCost with a dataloader, the indices parameter bypasses the dataloader and gathers data manually. From my experience dataloaders are slower when running on the cpu so its not commonly used but it might slow things down depending on the model and dataset size.

decent_bench/schemes.py
```diff
@@ -178,7 +178,7 @@ def __init__(self, n_significant_digits: int):
         self.n_significant_digits = n_significant_digits
 
     def compress(self, msg: Array) -> Array:  # noqa: D102
-        res = np.vectorize(lambda x: float(f"%.{self.n_significant_digits - 1}e" % x))(iop.to_numpy(msg))  # noqa: RUF073
+        res = np.vectorize(lambda x: float(format(x, f".{self.n_significant_digits - 1}e")))(iop.to_numpy(msg))
         return iop.to_array_like(res, msg)
```
Simpag, line 183: I feel like there has to be a more efficient way of performing quantization than doing to_numpy -> float -> string -> float -> back to framework. If you feel like you have the time please check if there are any better ways of doing this, otherwise I'll just create an issue of this at some point no problem.

Simpag (commented):
Two minor comments, dont have to do anything if you dont want to/have the time, otherwise LGTM.

decent_bench/schemes.py
Simpag: Update: You dont have to worry about this. This is insanely inefficient, I have made an update to this and will include it in my bigger update within 1-2 weeks. Some simple math made this at least 10x more efficient

adrianardv: Thanks, I’ll leave it as is now and rely on your upcoming update for the optimization

decent_bench/distributed_algorithms.py
```diff
@@ -369,11 +368,8 @@ def _run_minibatch_local_epochs(
         local_x: "Array",
     ) -> "Array":
         for _ in range(self.num_local_epochs):
-            indices = list(range(cost.n_samples))
-            random.shuffle(indices)
-            for start in range(0, cost.n_samples, cost.batch_size):
-                batch_indices = indices[start : start + cost.batch_size]
-                grad = cost.gradient(local_x, indices=batch_indices)
+            for _ in range(0, cost.n_samples, cost.batch_size):
```
nicola-bastianello: I only now realized this was doing a sweep across all data that belong to a client. this is aligned with the original FedAvg, but not with the more common implementation of it. in that, the clients do E epochs of SGD, where they use stochastic gradients with the default batch_size at each epoch of SGD. so I would remove the inner loop, just having `for _ in range(self.num_local_epochs): local_x -= self.step_size * cost.gradient(local_x)`

see Algorithm 1 in the FedProx paper

decent_bench/distributed_algorithms.py
```diff
@@ -323,44 +326,50 @@ def step(self, network: FedNetwork, iteration: int) -> None:  # noqa: D102
...
-        if isinstance(client.cost, EmpiricalRiskCost):
-            cost = client.cost
-            n_samples = cost.n_samples
-            return self._epoch_minibatch_update(cost, local_x, cost.batch_size, n_samples)
+        empirical_cost = self._empirical_cost_for_local_updates(client.cost)
+        if empirical_cost is not None:
+            return self._run_minibatch_local_epochs(empirical_cost, local_x)
+        return self._run_full_gradient_local_epochs(client.cost, local_x)
 
+    def _run_full_gradient_local_epochs(self, cost: Cost, local_x: "Array") -> "Array":
```
nicola-bastianello: if I'm not mistaken, with these change `_run_full_gradient_local_epochs` and `_run_minibatch_local_epochs` have the same steps. so we could simplify removing these two methods and just doing in `_compute_local_update`:
```
grad = cost.gradient(local_x)
local_x -= self.step_size * grad
```

adrianardv: Yes, with the code as it is now, it can be written like that, since both branches now perform the same kind of update. For empirical costs,  `gradient(local_x)` defaults to `indices="batch"`, while for plain Cost objects uses the implementation without batching. 

That said, I'm now thinking that now `_run_minibatch_local_epochs` , in each `self.num_local_epochs` performs the gradient over a single random minibatch, but epoch normally refers to a full pass over the dataset. So, the wording in this case is misleading.

image
Also, my understanding is that, in the FedAvg pseudocode in FedProx paper "E epoch of SGD" does means E full passes over the client's local data using minibatches.


So perhaps, if we keep the implementation as it is now, instead of talking about local epochs in the loop in  `_run_minibatch_local_epochs` maybe we should be talking about local steps? That being:
`local_steps=self.num_local_epochs * math.ceil(cost.n_samples / cost.batch_size)`

nicola-bastianello: you are right, they probably think of epoch as the full sweep over the whole dataset even in FedProx. I personally don't like fixing an algorithm to doing a full sweep during each local training, as it is limiting and in scenarios where the computational power is not a lot, the number of local steps could be limited.

also in my own papers I just think about steps, not a full sweep (and I'm not the only one)

so yes, you are right that they should be classified as steps rather than epochs. for the naming I would leave epoch for now, and we can decide after Simon has pushed some p2p algorithms that also use local training. we can then choose the name to use throughout so it's aligned

adrianardv: Okey, that makes sense. In that case, I'll apply the simplification you suggested, and keep the current epoch naming for now.

PR #284, opened 2026-04-02
feat(distributed_algorithms): Add FedProx algorithm

adrianardv:
This PR adds FedProx as a new federated algorithm. 

Implemented as a proper FedAlgorithm subclass that preserves the existing federated round structure used by FedAvg. The only behavioral difference is the local client update: each selected client optimizes the FedProx subproblem with a proximal term anchored at the round’s server model.

### DIFF
--- a/decent_bench/distributed_algorithms.py
+++ b/decent_bench/distributed_algorithms.py
@@ -451,7 +451,7 @@ class FedProx(FedAlgorithm):
         Costs that preserve the empirical-risk abstraction default ``gradient`` to ``indices="batch"``, so FedProx
         performs mini-batch local updates automatically. Generic costs keep their usual full-gradient behavior.
         """
-        reference_x = iop.copy(client.messages[server]) if server in client.messages else iop.copy(client.x)
+        reference_x = client.messages[server] if server in client.messages else client.x
         local_x = iop.copy(reference_x)
         for _ in range(self.num_local_epochs):
             grad = client.cost.gradient(local_x) + self.mu * (local_x - reference_x)
--- a/decent_bench/distributed_algorithms.py
+++ b/decent_bench/distributed_algorithms.py
@@ -370,16 +370,16 @@ class FedProx(FedAlgorithm):
     r"""
     Federated Proximal (FedProx) with local SGD epochs.
 
-    FedProx follows the same communication and aggregation pattern as
-    :class:`FedAvg <decent_bench.distributed_algorithms.FedAvg>`, but each client solves a proximalized local
-    subproblem around the round's server model:
+    Each client solves a proximalized local subproblem around the round's server model:
 
     .. math::
         h_k(\mathbf{w}; \mathbf{w}^t) = F_k(\mathbf{w}) + \frac{\mu}{2} \|\mathbf{w} - \mathbf{w}^t\|^2
 
     .. math::
-        \mathbf{x}_{i, k}^{(t+1)} = \mathbf{x}_{i, k}^{(t)} - \eta \left(
-        \nabla f_i(\mathbf{x}_{i, k}^{(t)}) + \mu (\mathbf{x}_{i, k}^{(t)} - \mathbf{w}^t) \right)
+        \mathbf{x}_{i, k}^{(t+1)} = \mathbf{x}_{i, k}^{(t)} - \eta
+        \nabla h_k(\mathbf{x}_{i, k}^{(t)}; \mathbf{w}^t)
+
+    where :math:`\nabla h_k(\mathbf{w}; \mathbf{w}^t) = \nabla F_k(\mathbf{w}) + \mu (\mathbf{w} - \mathbf{w}^t)`.
 
     .. math::
         \mathbf{x}_{k+1} = \frac{1}{|S_k|} \sum_{i \in S_k} \mathbf{x}_{i, k}^{(E)}
@@ -387,10 +387,9 @@ class FedProx(FedAlgorithm):
     where :math:`\mathbf{w}^t` is the server model broadcast at the start of round :math:`k`, held fixed
     throughout each selected client's local epochs, :math:`\mu \geq 0` is the proximal coefficient,
     :math:`\eta` is the step size, and :math:`S_k` is the set of participating clients. Setting ``mu=0.0``
-    recovers FedAvg exactly. The default ``mu=0.01`` is a sensible benchmark starting point, but users are
-    strongly encouraged to tune ``mu`` for each problem; a practical grid is ``[0.001, 0.01, 0.1, 0.5, 1.0]``.
-    As in FedAvg, aggregation uses client weights, defaulting to data-size weights when ``client_weights`` is not
-    provided, and client selection defaults to uniform sampling with fraction 1.0. For
+    recovers :class:`FedAvg <decent_bench.distributed_algorithms.FedAvg>` exactly. Aggregation uses client weights,
+    defaulting to data-size weights when ``client_weights`` is not provided, and client selection defaults to uniform
+    sampling with fraction 1.0. For
     :class:`~decent_bench.costs.EmpiricalRiskCost`, local updates use mini-batches of size
     :attr:`EmpiricalRiskCost.batch_size <decent_bench.costs.EmpiricalRiskCost.batch_size>`; for generic costs,
     local updates use full-batch gradients.
@@ -399,7 +398,7 @@ class FedProx(FedAlgorithm):
     iterations: int = 100
     step_size: float = 0.001
     num_local_epochs: int = 1
-    mu: float = 0.01  # Sensible benchmark default; tune over {0.001, 0.01, 0.1, 0.5, 1} for new settings.
+    mu: float = 0.01
     client_weights: ClientWeights | None = None
     selection_scheme: ClientSelectionScheme | None = field(
         default_factory=lambda: UniformClientSelection(client_fraction=1.0)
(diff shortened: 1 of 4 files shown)


decent_bench/distributed_algorithms.py
```diff
@@ -364,6 +364,118 @@ def _epoch_minibatch_update(
...
+@dataclass(eq=False)
+class FedProx(FedAlgorithm):
+    r"""
+    Federated Proximal (FedProx) with local SGD epochs.
+
+    FedProx follows the same communication and aggregation pattern as
+    :class:`FedAvg <decent_bench.distributed_algorithms.FedAvg>`, but each client solves a proximalized local
+    subproblem around the round's server model:
+
+    .. math::
```
nicola-bastianello, line 359: it might be better to write the local update in terms of h_k's gradient, since you have introduced the regularized cost. otherwise the notation is unused

decent_bench/distributed_algorithms.py
```diff
@@ -364,6 +364,118 @@ def _epoch_minibatch_update(
...
+        \nabla f_i(\mathbf{x}_{i, k}^{(t)}) + \mu (\mathbf{x}_{i, k}^{(t)} - \mathbf{w}^t) \right)
+
+    .. math::
+        \mathbf{x}_{k+1} = \frac{1}{|S_k|} \sum_{i \in S_k} \mathbf{x}_{i, k}^{(E)}
+
+    where :math:`\mathbf{w}^t` is the server model broadcast at the start of round :math:`k`, held fixed
+    throughout each selected client's local epochs, :math:`\mu \geq 0` is the proximal coefficient,
+    :math:`\eta` is the step size, and :math:`S_k` is the set of participating clients. Setting ``mu=0.0``
+    recovers FedAvg exactly. The default ``mu=0.01`` is a sensible benchmark starting point, but users are
+    strongly encouraged to tune ``mu`` for each problem; a practical grid is ``[0.001, 0.01, 0.1, 0.5, 1.0]``.
```
nicola-bastianello: I would remove the sentence `The default ``mu=0.01`` is a sensible benchmark starting point, but users are strongly encouraged to tune ``mu`` for each problem; a practical grid is ``[0.001, 0.01, 0.1, 0.5, 1.0]``.` this is because: that grid is potentially not the best starting point for some problems (as the best value of mu depends on the problem features). and we never mention tuning the hyperparameters, as it is implicit, so I don't think it's needed to mention it here

nicola-bastianello: I would also explicitly link FedAvg in the sentence `Setting ``mu=0.0`` recovers FedAvg exactly. `

decent_bench/distributed_algorithms.py
```diff
@@ -364,6 +364,118 @@ def _epoch_minibatch_update(
...
+    As in FedAvg, aggregation uses client weights, defaulting to data-size weights when ``client_weights`` is not
```
nicola-bastianello: I would not reference FedAvg in the docstring (except to say that it's the same as FedProx when mu=0). I think the docstring should self-contained

decent_bench/distributed_algorithms.py
```diff
@@ -364,6 +364,118 @@ def _epoch_minibatch_update(
...
+    provided, and client selection defaults to uniform sampling with fraction 1.0. For
+    :class:`~decent_bench.costs.EmpiricalRiskCost`, local updates use mini-batches of size
+    :attr:`EmpiricalRiskCost.batch_size <decent_bench.costs.EmpiricalRiskCost.batch_size>`; for generic costs,
+    local updates use full-batch gradients.
+    """
+
+    iterations: int = 100
+    step_size: float = 0.001
+    num_local_epochs: int = 1
+    mu: float = 0.01  # Sensible benchmark default; tune over {0.001, 0.01, 0.1, 0.5, 1} for new settings.
```
nicola-bastianello: I would remove the comment given the discussion above

docs/source/user.rst
```diff
@@ -108,6 +108,10 @@ Federated
             :tag: federated
             :module: decent_bench.distributed_algorithms
 
+FedProx extends FedAvg with a proximal coefficient ``mu``. Setting ``mu=0`` reduces
+FedProx to FedAvg. In practice, ``mu`` should typically be tuned for each problem;
```
nicola-bastianello: also here I would remove the discussion about tuning

decent_bench/distributed_algorithms.py
```diff
@@ -364,6 +364,118 @@ def _epoch_minibatch_update(
...
+        network.send(sender=network.server(), receiver=selected_clients, msg=network.server().x)
+
+    def _run_local_updates(self, network: FedNetwork, selected_clients: Sequence["Agent"]) -> None:
+        for client in selected_clients:
+            client.x = self._compute_local_update(client, network.server())
+            network.send(sender=client, receiver=network.server(), msg=client.x)
+
+    def _compute_local_update(self, client: "Agent", server: "Agent") -> "Array":
+        reference_x = iop.copy(client.messages[server]) if server in client.messages else iop.copy(client.x)
+        local_x = iop.copy(reference_x)
```
nicola-bastianello: if I'm not mistaken, this should not be here, because it effectively removes the regularization term local_x=reference_x. the result is that we always run FedAvg (I tested and that's indeed the case)

removing this line should be enough to fix it

nicola-bastianello (commented):
thanks! just a few minor comments, and one on the algorithm implementation

decent_bench/distributed_algorithms.py
nicola-bastianello: I've looked at the paper, and it's not very clear what they do to initialize `local_x`. I think maybe it should be left to the previous value? I tried the current implementation also with num local epochs > 1 and I get the same results for both algorithms. I don't quite understand

decent_bench/distributed_algorithms.py
```diff
@@ -364,6 +364,118 @@ def _epoch_minibatch_update(
...
+    def _sync_server_to_clients(self, network: FedNetwork, selected_clients: Sequence["Agent"]) -> None:
+        network.send(sender=network.server(), receiver=selected_clients, msg=network.server().x)
+
+    def _run_local_updates(self, network: FedNetwork, selected_clients: Sequence["Agent"]) -> None:
+        for client in selected_clients:
+            client.x = self._compute_local_update(client, network.server())
+            network.send(sender=client, receiver=network.server(), msg=client.x)
+
+    def _compute_local_update(self, client: "Agent", server: "Agent") -> "Array":
+        reference_x = iop.copy(client.messages[server]) if server in client.messages else iop.copy(client.x)
```
Simpag: You dont need to copy the reference_x since you're not modifying it

decent_bench/distributed_algorithms.py
```diff
@@ -364,6 +364,118 @@ def _epoch_minibatch_update(
...
+        per_client_batch: int,
+        n_samples: int,
+    ) -> "Array":
+        for _ in range(self.num_local_epochs):
+            indices = list(range(n_samples))
+            random.shuffle(indices)
+            for start in range(0, n_samples, per_client_batch):
+                batch_indices = indices[start : start + per_client_batch]
+                grad = cost.gradient(local_x, indices=batch_indices) + self.mu * (local_x - reference_x)
+                local_x -= self.step_size * grad
```
Simpag: If you want you could pass `indices="all"` here since the cost functions themselves should handle batching. But I'm not sure if that is what you want here so its just a comment

Simpag (commented):
Just some comments, no changes needed. LGTM

decent_bench/distributed_algorithms.py
adrianardv: > if I'm not mistaken, this should not be here, because it effectively removes the regularization term local_x=reference_x. the result is that we always run FedAvg (I tested and that's indeed the case)

My understanding is that setting local_x = reference_x removes the proximal term only at the first local step, since at that point local_x - reference_x = 0. After that, local_x is updated while reference_x stays fixed, so from the following local steps onward the proximal term becomes active again.

I took a closer look and ran a few local experiments to try to understand whether the current implementation was effectively collapsing to FedAvg. What I found was reassuring, although I’d definitely appreciate a second check on the interpretation.

I added a direct comparison plot of the server iterates, specifically the gap
`||x_server_prox - x_server_avg||` over the rounds.
This gap stays strictly above zero for all the mu values I tested, and it grows as mu increases, so the two methods do not seem to be following the exact same trajectory.

I also tracked the mean client drift from the server model. Here the behaviour looks quite consistent with what we would expect from FedProx: as mu increases, the average client drift decreases.

On top of that, I ran a few straggler-style experiments locally (0%, 50%, and 90% partial workers). In the 0% case the methods are indeed very close, which seems reasonable on a smooth quadratic problem. But in the 50% and especially 90% partial-worker settings, FedProx stays much more stable than FedAvg and the loss curves separate quite clearly.

So from these experiments, it looks like the current implementation is not collapsing to FedAvg.

> I've looked at the paper, and it's not very clear what they do to initialize local_x. I think maybe it should be left to the previous value?

Yes, the paper is not very explicit. I checked the original FedProx repository (https://github.com/litian96/FedProx), and there, before each selected client performs its local solve, the server calls c.set_params(self.latest_model), so local training is re-initialised from the current global model at the beginning of the round.

That was the reason I went with local_x = reference_x here, but I’d be very happy if you could double-check whether you think that is the right interpretation for this framework.

nicola-bastianello: ok, I can confirm that their behaviors do differ indeed. although this difference is not very big even when I select mu=100. but that's not on us, it's just the algorithm's design. I guess in my previous test I didn't go to values that were big enough to see a difference

and I agree with your interpretation that local_x = reference_x is the correct choice

nicola-bastianello:
looks good, thank you! merging now

PR #287, opened 2026-04-09
ref(centralized_algorithms): refactor solvers to classes and implement solve util

### LINKED ISSUE #177: Enhancing centralized_algorithms
### gradient_descent

If the cost function has `m_smooth` and `m_cvx` with `m_cvx`>0 we have an optimal step-size, which could be used by default (2/(m_smooth+m_cvx). And we could have a diminishing step-size for the case `m_cvx` = 0; this would be similar to the logic in `accelerated_gradient_descent`

### adding other algorithms

- proximal point algorithm (PPA)
- proximal gradient descent (for composite costs, receives 2 separate costs)
- Peaceman-Rachford splitting (for composite costs receives, 2 separate costs)

nicola-bastianello:
This PR refactors the centralized optimization solvers and updates related utilities. The main changes introduce a flexible, extensible solver framework with improved logging, parameter handling, and code clarity. The previous implementations of gradient descent and accelerated gradient descent are replaced by object-oriented solver classes, and a unified `solve` function is added to automatically select the appropriate optimization method. Logging and progress tracking are enhanced throughout. Additionally, the benchmark utility functions are updated to use the new solver interface and improved logging.

Key changes include:

### Centralized Solver Refactor and Enhancements

* Introduced a new object-oriented solver framework (`Solver`, `GradientDescent`, `AcceleratedGradientDescent` classes) replacing solver functions, providing extensibility and clearer parameter handling. The `run` method now manages iteration, stopping criteria, and progress display.
* Added a unified `solve` function that automatically selects the best solver based on the cost function's properties, improving usability and robustness.
* Improved logging and progress reporting using the `LOGGER` and `rich.progress.track`, and initialized logging at module load.

### Benchmark Utility Updates

* Updated `create_classification_problem`, `create_regression_problem`, and `create_quadratic_problem` in `_utils.py` to use the new `solve` function.

### Proximal Solver Improvements

* Refactored `proximal_solver` to use the new `AcceleratedGradientDescent` class, added stricter input validation, and improved error messages for unsupported cost types.

closes #177

### DIFF
--- a/decent_bench/centralized_algorithms.py
+++ b/decent_bench/centralized_algorithms.py
@@ -89,6 +89,7 @@ class Solver(ABC):
         if hyperparams:
             for k, v in hyperparams.items():
                 setattr(self, k, v)
+            self.hyperparams = hyperparams
 
     @abstractmethod
     def step(self, iteration: int) -> None:
@@ -109,6 +110,7 @@ class Solver(ABC):
         max_iter: int = 100,
         stop_tol: float | None = None,
         max_tol: float | None = None,
+        check_frequency: float = 0.01,
         show_progress: bool = True,
     ) -> Array:
         """
@@ -123,6 +125,9 @@ class Solver(ABC):
                       Must be positive if provided.
             max_tol: optional final tolerance; raises RuntimeError if ``norm(x_new - x_old) > max_tol`` after
                       max_iter iterations. Must be positive if provided.
+            check_frequency: float in (0, 1] defining how often the early stopping condition should be checked.
+                      A smaller value means that the stopping condition is checked more often.
+                      This applies only if ``stop_tol`` is not None.
             show_progress: whether to display a progress bar during iteration. Defaults to True.
 
         Returns:
@@ -142,6 +147,9 @@ class Solver(ABC):
             raise ValueError("`stop_tol` must be positive or None")
         if max_tol is not None and max_tol <= 0:
             raise ValueError("`max_tol` must be positive or None")
+        if check_frequency <= 0 or check_frequency > 1:
+            raise ValueError("`check_frequency` must be a float in (0, 1]")
+        check_every = max(1, int(check_frequency * max_iter))
 
         delta = np.inf
         for k in track(
@@ -152,9 +160,10 @@ class Solver(ABC):
         ):
             self.x_old = iop.copy(self.x)
             self.step(k)
-            delta = float(iop.norm(self.x - self.x_old))
-            if stop_tol is not None and delta <= stop_tol:
-                break
+            if stop_tol is not None and k % check_every == 0:
+                delta = float(iop.norm(self.x - self.x_old))
+                if delta <= stop_tol:
+                    break
 
         if max_tol is not None and delta > max_tol:
             raise RuntimeError(
@@ -178,24 +187,13 @@ class GradientDescent(Solver):
         if callable(step_size):
             step_size_k: Callable[[int], float] = step_size
         elif isinstance(step_size, float):
-
-            def step_size_k(_: int) -> float:
-                return float(step_size)
-
+            step_size_k = lambda _: float(step_size)  # noqa: E731
         elif np.isnan(cost.m_smooth) or np.isinf(cost.m_smooth) or np.isnan(cost.m_cvx):  # non-smooth or non-convex
-
-            def step_size_k(k: int) -> float:
-                return float(1 / np.sqrt(k + 1))
-
+            step_size_k = lambda k: float(1 / np.sqrt(k + 1))  # noqa: E731
         elif cost.m_cvx > 0:  # strongly convex
-
-            def step_size_k(_: int) -> float:
-                return 2 / (cost.m_smooth + cost.m_cvx)
-
+            step_size_k = lambda _: 2 / (cost.m_smooth + cost.m_cvx)  # noqa: E731
         else:  # convex
-
-            def step_size_k(_: int) -> float:
-                return 1 / cost.m_smooth
+            step_size_k = lambda _: 1 / cost.m_smooth  # noqa: E731
 
         super().__init__(cost, {"step_size": step_size_k}, x0)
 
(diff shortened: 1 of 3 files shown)


decent_bench/centralized_algorithms.py
```diff
@@ -1,137 +1,291 @@
...
+        if max_tol is not None and max_tol <= 0:
+            raise ValueError("`max_tol` must be positive or None")
+
+        delta = np.inf
+        for k in track(
+            range(max_iter),
+            description="Solving...",
+            disable=not show_progress,
+            update_period=0.0,
+        ):
```
Simpag, line 155: Could possibly set it to transient so the bar gets removed after it completes. Could also maybe dynamically disable/enable the progress bar and remove show_progress. Can be done using the following, or some variation of it:
```python
start_time = time.time()
bar_time_threshold = 2.0
bar_iter_threshold = 0.7 * max_iter
with Progress(transient=True, disable=True) as progress:
    task_id = progress.add_task("Solving...", total=max_iter)

    for k in range(max_iter):
        # Only enable prog bar if progress is less than bar_iter_threshold and time spent is more than bar_time_threshold
        if progress.disable and k < bar_iter_threshold and time.time() - start_time > bar_threshold:
            progress.disable = False
            progress.start()

        ...

        progress.advance(task_id, 1)
```

decent_bench/centralized_algorithms.py
```diff
@@ -1,137 +1,291 @@
...
+            self.x_old = iop.copy(self.x)
+            self.step(k)
+            delta = float(iop.norm(self.x - self.x_old))
```
Simpag: This delta operation could be somewhat expensive if x is high dimensional. Consider using the dot product instead (update target accordingly) and possibly only checking this every X iterations

decent_bench/centralized_algorithms.py
```diff
@@ -1,137 +1,291 @@
...
+            x0 = iop.zeros(shape=cost.shape, framework=cost.framework, device=cost.device)
+        if iop.shape(x0) != cost.shape:
+            raise ValueError("x0 and cost function domain must have same shape")
+        self.x = x0
+        self.x_old = iop.copy(self.x)
+
+        self.cost = cost
+        if hyperparams:
+            for k, v in hyperparams.items():
+                setattr(self, k, v)
```
Simpag: Concider also setting `self.hyperparams = hyperparams` as users might be confused if it doesnt exist and they dont know about this setattr

decent_bench/centralized_algorithms.py
```diff
@@ -1,137 +1,291 @@
...
+    def __init__(self, cost: "Cost", step_size: float | Callable[[int], float] | None = None, x0: Array | None = None):
+        if callable(step_size):
+            step_size_k: Callable[[int], float] = step_size
+        elif isinstance(step_size, float):
 
-    Raises:
-        RuntimeError: if ``norm(x_new - x) > max_tol`` at the end
-        ValueError: if ``cost.m_smooth < 0``, ``cost.m_cvx < 0``, or cost function is affine
+            def step_size_k(_: int) -> float:
+                return float(step_size)
```
Simpag: Could possibly do lambda functions instead to reduce empty space, or does mypy/ruff complain? If it doesnt work with lambdas you should reconsider the implementation as users will most likely use lambda functions as the step_size parameter if they want to use a function.

decent_bench/centralized_algorithms.py
```diff
@@ -1,137 +1,291 @@
...
+                    (np.sqrt(cost.m_smooth) - np.sqrt(cost.m_cvx)) / (np.sqrt(cost.m_smooth) + np.sqrt(cost.m_cvx))
+                )
+
+        else:
+
+            def momentum_k(k: int) -> float:
+                return k / (k + 3)
+
+        super().__init__(cost, {"step_size": step_size, "momentum": momentum_k}, x0)
+        self.y = iop.copy(self.x)
```
Simpag: Since you are using this approach here, setting self.y, it might be possible to remove the hyperparams argument from the base class. This would also allow implementations to call the super init as their first action

Simpag (commented):
Just some minor comments, overall good refactor

decent_bench/centralized_algorithms.py
nicola-bastianello: I'm not sure I understand, you mean doing 1) `delta = self.x.T @ self.x` or 2) `delta = (self.x-self.x_old).T @ (self.x-self.x_old)`? 1) will likely not work, as x is not guaranteed to be decreasing when the solver converges, only its distance from x_old. 2) I think should be as efficient as using `norm`, since I think `norm` is optimized

but yes, I agree that it is indeed an expensive operation. I'll implement a check every X iterations (I'm thinking X=max_iter/100). I'll also add an arg so that users can define this

decent_bench/centralized_algorithms.py
nicola-bastianello: yes I had to change to appease ruff; lambda functions passed by the user will work.

I will actually change it back to lambda and tell ruff to shut up, so that users can inspect the code and see lambdas

decent_bench/centralized_algorithms.py
nicola-bastianello: sorry I don't understand your suggestion, can you expand please

decent_bench/centralized_algorithms.py
nicola-bastianello: good point

decent_bench/centralized_algorithms.py
nicola-bastianello: I think I like seeing the progress bar even after it's complete, makes me feel like stuff is happening successfully. I think I also prefer to have show_progress for users to explicitly decide if the bar should be shown or not

decent_bench/centralized_algorithms.py
Simpag: I would do 
```python
delta = self.x - self.x_old
delta = delta.T @ delta
```
which should be significantly more efficient than using norm because it does not perform a sqrt operation, which is slow, but you could test it. It's not a huge deal since you won't be doing it as often as if you were working with graphics but its a common technique to use in those scenarios

decent_bench/centralized_algorithms.py
Simpag: You have the "hyperparams" argument in the base class but here you directly set `self.y = iop.copy(..)` after you call the superclass initialiser with step_size and momentum hyperparams. These hyperparams are just used to set the attribute values in the base class so the final result is the same as just setting "self.X = X" instead of passing it to the hyperparam argument. You should either 1) include "y" in the hyperparams passed to the superclass init, or 2) remove the hyperparam parameter and just make subclasses do manual assignment of the properties like you did with "y". I personally prefer option 2 as it is more clear and I don't think static type checkers are too happy with the attribute setting magic performed in the base class

decent_bench/centralized_algorithms.py
nicola-bastianello: I have the feeling the difference will be minimal since the sqrt is applied to a float, but I'll change it

decent_bench/centralized_algorithms.py
nicola-bastianello: ok, I agree it's cleaner to do 2). I wanted to reduce code repetition by using the superclass, but it's not a lot

nicola-bastianello: but I think I will still delegate initializing self.x to the superclass with its shape check

decent_bench/centralized_algorithms.py
Simpag: I believe the sqrt operation is about 50% of the total compute time in the norm method. However, the entire operation itself is fast so the total time difference is small. But its always nice to optimize

Simpag:
LGTM

PR #292, opened 2026-04-15
feat(scaffold): Add SCAFFOLD fed learning algorithm with control variates

adrianardv:
This PR adds a federated SCAFFOLD implementation and integrates it into the existing federated algorithm architecture.

The implementation follows the practical SCAFFOLD update with server and client control variates, preserves the current framework abstractions, and adds targeted tests.

## Main changes:
- Added `Scaffold` as a subclass of `FedAlgorithm` in `distributed_algorithms.py`.
- Implemented the practical SCAFFOLD control-variate update.
- Simplified built-in federated aggregation:
  - `FedAvg`, `FedProx`, and `Scaffold` now use uniform averaging.
  - removed configurable client-weight handling from the built-in federated algorithms.
- Moved the default uniform `aggregate(...)` implementation back into `FedAlgorithm`.
  - `FedAvg` and `FedProx` reuse the shared implementation.
  - `Scaffold` overrides it with its delta-based aggregation logic.
- Corrected federated participation semantics:
  - clients that do not receive the current server broadcast no longer fall back to their local state,
  - instead, only selected clients that actually receive the broadcast participate in the round.
- Added handling for buffered messages when `Network.buffer_messages=True`:
  - stale buffered client-to-server uploads are cleared before current-round local updates,
  - so aggregation uses only messages actually received in the current round.
- Corrected SCAFFOLD communication semantics so all round state goes through the network:
  - server -> client now sends a single message containing `(x, c)`,
  - client -> server now sends a single message containing `(delta_x, delta_c)`,
  - this ensures all SCAFFOLD communication is subject to drop / noise / compression consistently.
- Simplified SCAFFOLD auxiliary-variable handling:
  - persistent client state keeps only `c_i`,
  - round-specific server control `c` is taken from the received broadcast message,
- Updated user docs to match the simplified federated aggregation behavior.
- Added and updated tests covering:
  - SCAFFOLD control variates,
  - SCAFFOLD aggregation and partial participation,
  - dropped broadcasts,
  - buffered stale-message handling,
  - uniform federated aggregation behavior.


decent_bench/distributed_algorithms.py
```diff
@@ -432,14 +468,212 @@ def _compute_local_update(self, client: "Agent", server: "Agent") -> "Array":
...
+        self.selection_scheme = (
+            UniformClientSelection(client_fraction=1.0) if selection_scheme is None else selection_scheme
+        )
+        self.x0 = x0
+        self.name = name
+        self._use_uniform_client_weights_default = client_weights is self.CLIENT_WEIGHTS_UNSET
+        self.client_weights = (
+            None if self._use_uniform_client_weights_default else cast("ClientWeights | None", client_weights)
+        )
+        self.__post_init__()
```
Simpag: Maybe move this from `__init__` to either `__post_init__` or as first calls in `initialize`?

decent_bench/distributed_algorithms.py
```diff
@@ -432,14 +468,212 @@ def _compute_local_update(self, client: "Agent", server: "Agent") -> "Array":
...
             grad = client.cost.gradient(local_x) + self.mu * (local_x - reference_x)
             local_x -= self.step_size * grad
         return local_x
 
 
+@tags("federated")
+@dataclass(eq=False, init=False)
+class Scaffold(FedAlgorithm):
+    r"""
+    SCAFFOLD with client/server control variates for variance-reduced local training.
```
Simpag: Add a reference to the paper if applicable, you would need to add it to the references.bib in the docs and add a `.. footbibliography::` at the end of the doc-string. See ATC for an example

Simpag (commented):
Minor comments regarding use of `__init__` and just citations. Otherwise looks good

decent_bench/distributed_algorithms.py
```diff
@@ -172,7 +172,13 @@ class FedAlgorithm(Algorithm[FedNetwork]):
     selection_scheme: ClientSelectionScheme | None = None
     _DEFAULT_SELECTION_SCHEME: Final[object] = object()
     client_weights: ClientWeights | None = None
-    _DEFAULT_CLIENT_WEIGHTS: Final[object] = object()
+    """Server aggregation weights.
+
+    If ``aggregate(..., client_weights=...)`` is given an explicit value, that value is used for that call.
+    Otherwise ``self.client_weights`` is used. If the resolved value is ``None``, weights are inferred from client
+    data size via :func:`decent_bench.utils.algorithm_helpers.infer_client_weight`.
```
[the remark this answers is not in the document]
Simpag: Im not sure if these Copilot comments make sense. You can probably ignore them unless you understand what it means

nicola-bastianello:
thank you! I have quite a lot of significant comments, but the algorithm implementation looks good

I've been thinking a lot about the role of the algorithms we provide in the code base. my conclusion is: the algorithms should be a pedagogical tool (to teach users how to code their own), as well as of course a convenience so that users don't have to code common algorithms

based on this observation, I have some general comments regarding the current implementation of the federated algorithms:

1. I would remove the ClientWeights type, and only allow `dict[Agent, float] | None`. this would align it to the types we pass e.g. for network schemes. I would in general avoid lists for client-specific attributes
2. I find scaffold's weights especially confusing because in the docs we get `client_weights: ClientWeights | object | None = FedAlgorithm.CLIENT_WEIGHTS_UNSET`, where `FedAlgorithm.CLIENT_WEIGHTS_UNSET` is not a hyperlink. so it's difficult to get a sense for what this means without reading the docstring and the user guide
3. also, I would argue that we should not allow to override scaffold's weights, since this is part of how the algorithm is defined (\Delta c = 1/|S_k| \sum_{i \in S_k} \Delta c_i). this is not something we can allow to change, otherwise the whole algorithm can diverge or misbehave
4. for FedAvg: the docstring shows 1/|S_k| \sum_{i \in S_k}, but the implementation defaults to data-based weights. I would argue that we should use 1/|S_k| \sum_{i \in S_k} always, without allowing to set the weights. this simplifies understanding the code quite a lot and aligns the algorithm with the docstring. if users want special weights, they either multiply the costs or define a new algorithm. so also for FedAvg I would remove the client weights option
5. FedProx (like scaffold) defines specific weights in the paper: 1/|S_k| \sum_{i \in S_k}. so this is part of the algorithm and we should not allow the users to change it

basically, my idea is that from a pedagogical/user-friendliness perspective, we should not allow setting the weights. we just present the updates with uniform weights in the docstring, and have the implementation match that. of course, you keep the more advanced versions in your experiments for the thesis, but they are not part of the codebase default algorithms

other comments:

- scaffold's updates in the docstring are incomplete (definition of Delta c, Delta x are missing)
- eta_l, eta_g are not defined in the docstring
- removing the client weights would also likely significantly simplify the private methods in FedAlgorithm; currently, I have to say I find the implementation confusing
- in general, I pesonally find quite difficult to read algorithms where we have several nested private methods, which seems to be your style. it's not a hard requirement to change this to a more linear structure (like the algorithms for P2P), but I think it would help make them pedagogical tools
- I'm not sure I see the need for the three `ClassVar` keys definitions. I would simply define these variables in `initialize`, like we do for all auxiliary variables in other algorithms. I would also rename them to `c`, `c_i`, `delta_c` etc, to match the docstring. this is how we are doing it in other algorithms. but maybe I'm missing something, so please let me know if that's the case

decent_bench/distributed_algorithms.py
nicola-bastianello: I agree, let's use a uniform implementation style throughout: args validation in `__post_init__` and variables/auxiliary variables definitions in `initialize`

this maybe was due to having customizable client weights, but given my comment below customizing the init should not be needed

decent_bench/distributed_algorithms.py
nicola-bastianello: and also in FedAvg and FedProx please, I didn't notice they were missing

adrianardv:
Thanks, this is very helpful.

I agree with your point on SCAFFOLD. There, the aggregation is part of the algorithm definition, so it makes sense not to expose client weights and to match the paper pseudocode exactly, i.e. uniform averaging over the selected clients. I will change that.

For FedAvg and FedProx, the reason I initially introduced configurable client weights is that, when looking at the literature and existing implementations, there seems to be some inconsistency in how the aggregation is defined.

In the FedAvg paper, the aggregation is data-size weighted. In contrast, in the FedProx paper, both the FedAvg algorithm they restate and the FedProx pseudocode are written with uniform averaging over the selected clients. At the same time, in the official FedProx repository, both FedAvg and FedProx use data-size-weighted aggregation. So it seems that both conventions are used in practice.

My initial idea was therefore to expose this flexibility so that users could choose between these options without having to implement a new algorithm.

That said, I understand your point regarding simplicity and readability. I agree that the current ClientWeights abstraction makes the implementation harder to follow.

As a compromise, I’m thinking of simplifying this by only allowing two explicit options for FedAvg and FedProx:
- uniform averaging
- data-size-weighted averaging

with uniform averaging as the default, and no support for arbitrary custom weights. This would keep the implementation simple, while still covering the two common variants used in the literature and in practice.

Let me know what you think about this approach.

I will also go through the other points you mention, and fix/improve those things.

nicola-bastianello:
thanks for the explanation Adriana, I see that they made a mess with these weights between papers and code.

I was thinking about proposing the same solution of allowing uniform or data-based only. but I decided against it because it would complicate the docstring (not by a lot, but still), and more importantly because from a theoretical standpoint I don't agree with the use of data-based weights in the aggregation if they are not also part of the cost definition (that is, we are solving sum_i w_i f_i). if we have data-based weights in the aggregate, but are solving \sum_i f_i / N, then using a weighted aggregate actually goes against finding a stationary point of the problem. sure, it might help with accuracy (which is why I guess they secretly use it in the code but don't show it in the paper), but it is not theoretically sound. and I prefer the codebase algorithms to be theoretically rigorous

one could allow providing custom weights (or only data-based/uniform) for the cost, and then use those in the aggregate of the algorithms. but this would mean quite a lot of custom logic that would be difficult to understand. so overall, I would completely remove weights to simplify and ensure theoretical rigor.

but in your thesis I think it will be nice to compare the two versions of the algorithms, with and without data-based weights. especially it will be interesting to see how optimality and accuracy change if weights are used or not

decent_bench/distributed_algorithms.py
```diff
@@ -341,18 +271,32 @@ def _compute_local_update(self, client: "Agent", server: "Agent") -> "Array":
...
         performs mini-batch local updates automatically. Generic costs keep their usual full-gradient behavior.
         """
-        local_x = iop.copy(client.messages[server]) if server in client.messages else iop.copy(client.x)
+        local_x = self._client_reference_x(client, server)
         for _ in range(self.num_local_epochs):
             grad = client.cost.gradient(local_x)
             local_x -= self.step_size * grad
         return local_x
 
+    def aggregate(
```
nicola-bastianello: I think I liked having `aggregate` as a default method in the superclass; Scaffold can override it, but the others can keep it as it is

decent_bench/distributed_algorithms.py
```diff
@@ -181,96 +170,36 @@ def server_broadcast(self, network: FedNetwork, selected_clients: Sequence["Agen
...
-    def select_clients(
-        self,
-        clients: Sequence["Agent"],
-        iteration: int,
-        selection_scheme: ClientSelectionScheme | object | None = _DEFAULT_SELECTION_SCHEME,
-    ) -> list["Agent"]:
-        """
-        Select participating clients from an eligible pool.
+    @staticmethod
+    def _client_reference_x(client: "Agent", server: "Agent") -> "Array":
```
nicola-bastianello: I've been thinking about this: 1) the name is, to me, a bit uninformative. maybe something like `_get_server_broadcast` might work better; 2) please correct me if I'm wrong, but I see a problem with using the client's x as fallback: it modifies the algorithm. if a client is active, it does not receive server.x, and falls back on its own state: this means that it will perform essentially 2*num_local_epochs local training steps. but this is a non-trivial modification of the algorithm. when implementing the p2p ones we debated this, and we decided not to modify the logic of the algorithms beyond making it work in all scenarios. so in this case, what I would do is to deactivate a client that does not receive server.x. probably we then need to use the intersection of selected clients and active clients for aggregate

decent_bench/distributed_algorithms.py
```diff
@@ -432,13 +376,181 @@ def _compute_local_update(self, client: "Agent", server: "Agent") -> "Array":
...
+        self.x0 = alg_helpers.initial_states(self.x0, network)
+        server = network.server()
+        server_x0 = self.x0[server]
+        server.initialize(
+            x=server_x0,
+            aux_vars={"c": iop.zeros_like(server_x0)},
+        )
+        for client in network.clients():
+            client_x0 = self.x0[client]
+            zero_control = iop.zeros_like(client_x0)
```
nicola-bastianello: c_i and c are both initialized as `iop.zeros_like(client_x0)`, so you could define zero_control like now, and then `iop.copy` it to define c_i and c

nicola-bastianello: or probably it's more efficient to just do zeros_like in-place instead of copying zero_control twice

decent_bench/distributed_algorithms.py
```diff
@@ -432,13 +376,181 @@ def _compute_local_update(self, client: "Agent", server: "Agent") -> "Array":
...
+    def step(self, network: FedNetwork, iteration: int) -> None:  # noqa: D102
+        selected_clients = self._selected_clients_for_round(network, iteration)
+        if not selected_clients:
+            return
+
+        self.server_broadcast(network, selected_clients)
+        server_control = network.server().aux_vars["c"]
+        for client in selected_clients:
+            if network.server() in client.messages:
+                client.aux_vars["c"] = iop.copy(server_control)
```
nicola-bastianello: if I don't misunderstand, copying `server_control` here means that this doesn't actually go through the network as it should. instead, this should be sent as a message. I think you could do this by stacking server.x and server.aux_vars["c"] and send that. this way they are correctly subjected to possible loss, noise, compression etc

nicola-bastianello (commented):
a few more comments but looks good overall, thank you!

decent_bench/distributed_algorithms.py
adrianardv: Yes, you're right.
I changed the scaffold broadcast to send a single stacked message containing both `(x,c)`. 
And, after local training, clients send back a single stacked message containing `(delta_x, delta_c)`

decent_bench/distributed_algorithms.py
adrianardv: Yes, you're right. I changed this.

- Renamed the helper to _get_server_broadcast
- Removed the fallback to the client's local x. Now:
1. the server broadcasts to selected clients
2. i then defined the actual participating clients as the subset of selected clients that actually received the broadcast.
3. only those clients run local training and send the updated x back to the server.
4. aggregation is performed over clients that participated and whose uploads were actually received by the server.

nicola-bastianello:
looks good, thank you! merging now

PR #294, opened 2026-04-22
feat(distributed_algorithms): Add FedOpt federated variants

adrianardv:
This PR adds FedOpt support to the federated algorithms by introducing
`FedAdagrad`, `FedYogi`, and `FedAdam`.

## Changes

- add a shared `FedOpt` base class
- add `FedAdagrad`, `FedYogi`, and `FedAdam`
- keep aggregation uniform across participating clients only
- add FedOpt docstrings and paper citations
- add tests


nicola-bastianello:
looks good, thank you! the only comment I have: it would be better to show the updates m_t and x_t on the server side also in FedOpt, with the definitions of the hyperparameters tau, beta_1, eta. the update v_t can be written as v_t = \text{to be implemented in subclasses} or something like that

adrianardv:
Just a small note regarding the aggregation choice in FedOpt.

Following our discussion in #292, I implemented FedOpt using uniform aggregation.

However, for completeness, I wanted to point out that in the FedOpt paper there is also some inconsistency similar to what we saw:
- in the main text  (algorithms 1 and 2), the aggregation is written as uniform average over the selected clients.
- In the appendix (Algorithm 5), the version used for implementation, switches to a data-size-weighted aggregation.
- They also show scaffold algorithm in the appendix (algorithm 6) using data-size-weighted aggregation.
- Also, the official FedOpt repository also appears to use data-size-weighted aggregation.

nicola-bastianello:
I see, thank you for the clarification. I see that they make a huge mess about this. the underlying reason, I think, is that the specific choice of aggregation for all these algorithms (except scaffold) is a heuristic. that is: when num_local_epochs>1, none of these algorithms guarantee convergence to a optimal/stationary point (due to client drift). so the choice of aggregation is something one can do to try and improve results in practice. and I'm still of the opinion that multiplying the client losses by data-based weights makes more sense

nicola-bastianello:
anyway, looks good thank you! merging now

PR #298, opened 2026-04-24
feat(distributed_algorithms): Add FedNova 

adrianardv:
Implementation of FedNova, including:
- plan no-momentum FedNova
- data-proportional client weighting following the paper and official github
- heterogeneous per-client local step counts
- optional local momentum
- optional proximal local updates
- optional server momentum
Default behaviour is plain FedNova: no local momentum, no proximal term, no server momentum.

Added tests for FedNova:
- equivalence to FedAvg when local steps are equal
- heterogeneous local steps handling
- etc.


decent_bench/utils/types.py
```diff
@@ -14,6 +14,8 @@
     from decent_bench.agents import Agent
     from decent_bench.networks import Network
     from decent_bench.utils.array import Array
+else:
+    Agent = Any
```
nicola-bastianello: why do this?

decent_bench/utils/types.py
```diff
@@ -41,6 +43,14 @@
 :func:`~decent_bench.utils.algorithm_helpers.initial_states`.
 """
 
+type LocalSteps = int | dict[Agent, int]
```
nicola-bastianello: since it's only the union of two things, and it's used only in one place, I would remove this custom type to make reading the docs more immediate

decent_bench/distributed_algorithms.py
```diff
@@ -761,6 +763,320 @@ def _update_second_moment(self, second_moment: "Array", average_delta: "Array")
...
+        \mathbf{v}_{i, t}^{(k+1)} = \beta \mathbf{v}_{i, t}^{(k)} + \mathbf{g}_{i, t}^{(k)},
+        \qquad
+        \mathbf{d}_{i, t}^{(k)} = \mathbf{v}_{i, t}^{(k+1)},
+
+    otherwise :math:`\mathbf{d}_{i, t}^{(k)} = \mathbf{g}_{i, t}^{(k)}`. The local model update is
+
+    .. math::
+        \mathbf{x}_{i, t}^{(k+1)} = \mathbf{x}_{i, t}^{(k)} - \eta_l \mathbf{d}_{i, t}^{(k)}.
+
+    Client :math:`i` accumulates the communicated update
```
nicola-bastianello: why "communicated"? at this stage nothing has been communicated if I'm not mistaken

decent_bench/distributed_algorithms.py
```diff
@@ -761,6 +763,320 @@ def _update_second_moment(self, second_moment: "Array", average_delta: "Array")
...
+
+    .. math::
+        \mathbf{c}_i^t = \sum_{k=0}^{\tau_i - 1} \eta_l \mathbf{d}_{i, t}^{(k)},
+
+    and maintains the FedNova scalar recurrences
+
+    .. math::
+        s_i^{(k+1)} = \beta s_i^{(k)} + 1
+
+    when ``use_momentum=True`` and :math:`s_i^{(k+1)} = 1` otherwise, together with
```
nicola-bastianello: ``use_momentum=True``__,__ and :math:`s_i^{(k+1)} = 1`

nicola-bastianello: comma after use_momentum=True

decent_bench/distributed_algorithms.py
```diff
@@ -761,6 +763,320 @@ def _update_second_moment(self, second_moment: "Array", average_delta: "Array")
...
+
+    .. math::
+        a_i^{(k+1)} = (1 - \eta_l \mu) a_i^{(k)} + s_i^{(k+1)}
+
+    when ``use_prox=True`` and :math:`a_i^{(k+1)} = a_i^{(k)} + s_i^{(k+1)}` otherwise.
+    After :math:`\tau_i` local steps, client :math:`i` uploads :math:`(\mathbf{c}_i^t, a_i, p_i)`, where the
+    data-proportional client weight is
```
nicola-bastianello: this weight needs information about the number of datapoints of other clients, which can only come from the server. I would specify that this is the case (e.g. say at the beginning the server collects n_i from all clients and sends p_i back)

nicola-bastianello: also, I didn't see a definition for n_i

decent_bench/distributed_algorithms.py
```diff
@@ -761,6 +763,320 @@ def _update_second_moment(self, second_moment: "Array", average_delta: "Array")
...
+    (single int ``local_steps``) or heterogeneous via a mapping keyed by client agent. In this implementation,
+    :math:`n_i` is inferred from each participating client's local cost via
+    :func:`~decent_bench.utils.algorithm_helpers.infer_client_weight`.
+
+    .. footbibliography::
+    """
+
+    iterations: int = 100
+    step_size: float = 0.001
+    local_steps: LocalSteps = 1
```
nicola-bastianello: I would call this num_local_steps (or epochs, and I'll change to steps later)

decent_bench/distributed_algorithms.py
```diff
@@ -761,6 +763,320 @@ def _update_second_moment(self, second_moment: "Array", average_delta: "Array")
...
+    use_momentum: bool = False
+    use_prox: bool = False
+    use_server_momentum: bool = False
+    beta: float = 0.9
+    mu: float = 0.01
+    gamma: float = 0.9
```
nicola-bastianello, line 858: it might be good to arrange these args as: use_momentum, beta; use_prox, mu; use_server_momentum, gamma

this way the use_ arg is paired with the corresponding hyperparameter

decent_bench/distributed_algorithms.py
```diff
@@ -761,6 +763,320 @@ def _update_second_moment(self, second_moment: "Array", average_delta: "Array")
...
+            raise ValueError("`step_size` must be positive")
+        if not (0 <= self.beta < 1):
+            raise ValueError("`beta` must satisfy 0 <= beta < 1")
+        if self.mu < 0:
+            raise ValueError("`mu` must be non-negative")
+        if not (0 <= self.gamma < 1):
+            raise ValueError("`gamma` must satisfy 0 <= gamma < 1")
+        self.local_steps = self._normalize_local_steps_spec()
+
+    @staticmethod
```
nicola-bastianello: it seems to me that the treatment of local steps is quite confused and inefficient. the role of _local_steps_by_client is also not clear to me.

how I would do it (in a single private method, or even directly in post_init):

1. if local_steps is int, validate it's >0, and replace it by {client: local_steps for client in network.clients}
2. if local_steps is dict, validate each element and verify that all clients have a value. if you want to silently ignore keys that are not in there, you can create a new dict that overwrites self.local_steps (see how message scheme dicts are handled)

I think this would make for much neater code

decent_bench/distributed_algorithms.py
```diff
@@ -761,6 +763,320 @@ def _update_second_moment(self, second_moment: "Array", average_delta: "Array")
...
+        participating_clients = self._clients_with_server_broadcast(network, selected_clients)
+        if not participating_clients:
+            return
+        self._clear_buffered_server_messages(network, participating_clients)
+        self._run_local_updates(network, participating_clients)
+        self.aggregate(network, participating_clients)
+
+    def _run_local_updates(self, network: FedNetwork, participating_clients: Sequence["Agent"]) -> None:
+        server = network.server()
+        raw_weights = [alg_helpers.infer_client_weight(client) for client in participating_clients]
```
nicola-bastianello: why is this performed here? if I don't misunderstand the algorithm, client weights p_i don't change over time, so the definition of client_weights can happen in initialize.

I would also do the following: the weights are defined in initialize and then p_i is stored in the client, and the server stores the whole dict of p_i (if needed, I'm not sure it is but the algorithm description in the paper is unreadable); then agents send a_i and d_i already multiplied by p_i

decent_bench/distributed_algorithms.py
```diff
@@ -761,6 +763,320 @@ def _update_second_moment(self, second_moment: "Array", average_delta: "Array")
...
+
+        Raises:
+            RuntimeError: if local steps have not been initialized for ``client``.
+
+        """
+        reference_x = self._get_server_broadcast(client, server)
+        local_x = iop.copy(reference_x)
+        cumulative_gradient = iop.zeros_like(reference_x)
+        local_momentum = iop.zeros_like(reference_x)
+        if client not in self._local_steps_by_client:
```
nicola-bastianello: this check can be removed after the init of local steps is refactored

decent_bench/distributed_algorithms.py
```diff
@@ -761,6 +763,320 @@ def _update_second_moment(self, second_moment: "Array", average_delta: "Array")
...
+
+            if self.use_prox:
+                a_i = ((1 - (self.step_size * self.mu)) * a_i) + momentum_scalar
+            else:
+                a_i += momentum_scalar
+
+        return local_x, cumulative_gradient, a_i
+
+    @staticmethod
+    def _pack_local_update_payload(cumulative_gradient: "Array", a_i: float, client_weight: float) -> "Array":
```
nicola-bastianello: this is a confusing name in my opinion. I would maybe call it _prepare_upload or something like that

also, performing to_numpy and flattening every time is likely super inefficient. what I would do is a sequence of uploads to the server, one for each of the quantities that need to be transmitted. the packet loss is fixed to the iteration number, so if it is true, all of the messages will be lost, which is kind of the same as sending one packet that is lost

decent_bench/distributed_algorithms.py
```diff
@@ -761,6 +763,320 @@ def _update_second_moment(self, second_moment: "Array", average_delta: "Array")
...
+        server_x = iop.copy(server.x)
+        decoded_uploads = [
+            self._unpack_local_update_payload(server.messages[client], iop.shape(server_x))
+            for client in received_clients
+        ]
+        cumulative_gradients = [upload[0] for upload in decoded_uploads]
+        a_values = [upload[1] for upload in decoded_uploads]
+        client_weights = [upload[2] for upload in decoded_uploads]
+
+        if any(a_i <= 0 for a_i in a_values):
```
nicola-bastianello, line 1043: is there any way that this is not true? to me it looks like they must be positive by construction, so I woudl remove the check

nicola-bastianello (commented):
looks good, thank you! but there are several parts that in my opinion could be improved

decent_bench/utils/types.py
adrianardv: I had to move it there because sphinx was failing,  but I'll try again.

I've been trying and I haven't managed to make it work without the custom type without sphinx crashing.

decent_bench/utils/types.py
adrianardv: I had to do that because `Agent` had to exist there for the type `LocalSteps = int | dict[Agent, int]` to work. Having `"Agent"` inside the dict, that fallback is not needed, but then ruff fails.

decent_bench/distributed_algorithms.py
adrianardv: you're right, I'll reword it

decent_bench/distributed_algorithms.py
adrianardv: Yeah, I see.
I actually did that because, if i'm not wrong, in the current network desing, each receiver stores only one message per sender, so sending `msg1`, `msg2`, and `msg3` separatelly would make later sends overwrite earlier ones. So, when in aggregate the server recollects the messages it will only get `msg3`.
Also, the drop decision is currently made on each individual send() call, it is not fixed per iteration.

---

I've now changed it to:
A message can now be either a single array or a grouped upload packet (`dict[str, Array]`).
The network treats it a sone logical upload: one stored message per sender, one drop decision.
If the message is grouped, compression and noise are applied to each array field separately.

decent_bench/distributed_algorithms.py
adrianardv: Good point. Okey, changed it to:
- In `__post_init__`, validate: 
     - if it is a scalar check >0. 
     - If it is a dict, check every value is >0. 
     
     (Here only num validation check because the network is still not available)
- In `initialize()`: 
     - if it is an int:  {client: num_local_steps for client in network.clients()}
     - if it is already a dict: check that every network client has a value, and  extra keys are ignored

decent_bench/distributed_algorithms.py
adrianardv: nice, moved it to initialize.

If we send d_i, we need to do the step size multiplication on the server side.

Yeah, what they do in the algorithm description in the paper is:
They compute p_i over all clients. But then in each round only a subset of clients participate, so the server renormalises the weights over the participating set.
In that algorithm they do not specify the weights handling. It just seems like the server knows about the weights. 
But then in the second algorithm, they say clients send piai and pidi.

I checked the offcial github and there they do a simplification:
server knows all n_i, and computes p_i s at initialisation. Then they directly use those p_i without renormalisation.
Also, the paper mathematical/abstract definition they do of di, is different  from the practical way they do it in the repository (with cum_grad as I've done it too).

adrianardv:
Regarding your comments about where the weights live, and what clients communicate.
I've been thinking about it, and how they do it in the paper and github.
And, I've finally adapted the code to this:

- at the start, the server stores each clients n_i.
- then in each round, the server looks at which clients actually participated and computes the weights p_i over that round's participating set. 
So the weights are always normalised for the current round, which matches the FedNova formulation.

- clients send a_i and cum_grad_i.
- they do not send p_i or n_i.
Responsabilities are clean: clients compute local stuff, server computes global aggregation.


Let me know what you think

decent_bench/utils/types.py
nicola-bastianello: ok, I really hate when this happens. for now leave it like it is, I'll come back to it at some point to try and remove as many custom types (I consider a custom type redundant if it is the union of only 2 things and is not used many times)

decent_bench/networks.py
```diff
@@ -236,23 +237,30 @@ def active_connected_agents(self, agent: Agent) -> list[Agent]:
...
 
         The message will be immediately available to the receiver if it is active in the current iteration.
         """
         sender._n_sent_messages += 1  # noqa: SLF001
         if self._message_drop[sender].should_drop():
             sender._n_sent_messages_dropped += 1  # noqa: SLF001
             return
-        msg = self._message_compression[sender].compress(msg)
-        msg = self._message_noise[sender].make_noise(msg)
+        if isinstance(msg, dict):
```
nicola-bastianello: I like the idea of being able to send dicts, but I'm afraid this might have a lot of unforeseen consequences. for example, in many places we assume the message to be an Array that we can do iop.shape() and other operations on. I also don't like having custom type `UploadPacket` nested into another custom type `NetworkMessage`, makes the docs more difficult to read

for now, I suggest removing the use of dict messages. instead, I would do separate transmissions (1 for loop of client transmission + server aggregation for each message). I know the could will become a bit less linear, but I don't think we can afford to change such a fundamental assumption (msg is Array) right now

let's open an issue to discuss how best to do this. but I think we will need to revisit it after the iop refactor, possibly adding a new pathway in the iop Backend that can handle dicts and apply the required iop operation to each of the values in the dict

decent_bench/distributed_algorithms.py
```diff
@@ -761,6 +789,309 @@ def _update_second_moment(self, second_moment: "Array", average_delta: "Array")
...
+
+    def _validate_num_local_steps(self) -> None:
+        if isinstance(self.num_local_steps, int):
+            if self.num_local_steps <= 0:
+                raise ValueError("`num_local_steps` must be positive")
+            return
+        if isinstance(self.num_local_steps, dict):
+            for step in self.num_local_steps.values():
+                if step <= 0:
+                    raise ValueError("`num_local_steps` must be positive")
```
nicola-bastianello: maybe "must have positive values" since in this branch it's a dict

decent_bench/distributed_algorithms.py
```diff
@@ -761,6 +789,309 @@ def _update_second_moment(self, second_moment: "Array", average_delta: "Array")
...
+                "`num_local_steps` mapping must provide a value for every network client; "
+                f"missing clients: {missing_clients}"
+            )
+        return {client: self.num_local_steps[client] for client in clients}
+
+    def _resolve_client_sample_counts(self, network: FedNetwork) -> dict["Agent", float]:
+        client_sample_counts: dict[Agent, float] = {}
+        for client in network.clients():
+            client_sample_count = alg_helpers.infer_client_weight(client)
+            if client_sample_count <= 0:
```
nicola-bastianello: I don't see how the client_sample_count could ever be <=0. if I'm not mistaken infer_client_weight can only output a positive float or raise

decent_bench/distributed_algorithms.py
```diff
@@ -761,6 +789,309 @@ def _update_second_moment(self, second_moment: "Array", average_delta: "Array")
...
+                raise ValueError("FedNova client sample counts must be positive")
+            client_sample_counts[client] = client_sample_count
+        return client_sample_counts
+
+    def _server_client_sample_counts(self, server: "Agent") -> dict["Agent", float]:
+        client_sample_counts = server.aux_vars.get("client_sample_counts")
+        if client_sample_counts is None:
```
nicola-bastianello: can this ever happen? I think not, since this is hard-coded in the initialize

decent_bench/distributed_algorithms.py
```diff
@@ -761,6 +789,309 @@ def _update_second_moment(self, second_moment: "Array", average_delta: "Array")
...
+            raise RuntimeError("FedNova server must be initialized with client sample counts before aggregation")
+        if not isinstance(client_sample_counts, dict):
+            raise TypeError("FedNova server client sample counts must be stored as a dict")
+        return cast("dict[Agent, float]", client_sample_counts)
+
+    def _client_num_local_steps(self, client: "Agent") -> int:
```
nicola-bastianello: I'm not sure I see the purpose of this function. once the num_local_steps arg has been validated and passed through `_settle_num_local_steps` it will be correct (each client has one number of local steps), and then it can simply be accessed as a normal dict

if I'm not mistaken, every other path has already raised

decent_bench/networks.py
nicola-bastianello: also, having to explicitly do `_require_array_message` in most algorithms is not a good design choice. this should happen in the backend, if we decide to go for implementing dict messages after the iop refactor

nicola-bastianello (commented):
thanks for the revision! there are still a few issues though

decent_bench/distributed_algorithms.py
nicola-bastianello: see my comment on the risk of allowing dict messages. I would write the algorithm with a sequence of client uploads and server aggregation on each of the transmitted messages (repetitive, I know, but we can fix it in the future).

for the drop, for now let's just accept that only a subset of the messages might be dropped (instead of all/none). we can address this also in the future

nicola-bastianello:
> Regarding your comments about where the weights live, and what clients communicate. I've been thinking about it, and how they do it in the paper and github. And, I've finally adapted the code to this:
> 
>     * at the start, the server stores each clients n_i.
> 
>     * then in each round, the server looks at which clients actually participated and computes the weights p_i over that round's participating set.
>       So the weights are always normalised for the current round, which matches the FedNova formulation.
> 
>     * clients send a_i and cum_grad_i.
> 
>     * they do not send p_i or n_i.
>       Responsabilities are clean: clients compute local stuff, server computes global aggregation.
> 
> 
> Let me know what you think

yes, I think this is the best way to do it, thank you!

decent_bench/networks.py
adrianardv: Yes, it would be good to be able to send dicts, but rn we have the core assumption that msg is an Array throughout the code. That's kind of why I had to introduce _require_array_message; because once I changed the shared message type to allow dicts, mypy could no longer tell that most existing algoritms were still receiving plain arrays, so I had to narrow those reads explicitly. So, there should be a better way to do this. 

Okey, I'll open the issue and do the separate transmissions you propose in FedNova.

decent_bench/utils/algorithm_helpers.py
```diff
@@ -157,31 +157,38 @@ def infer_client_weight(client: "Agent") -> float:
...
+    as ``A`` or ``b`` on the cost.
 
     Raises:
-        ValueError: if a suitable size attribute is not found on the cost.
+        ValueError: if a suitable positive size attribute is not found on the cost.
 
     """
+
+    def _validate_size(size: float) -> float:
+        if size <= 0:
```
nicola-bastianello, line 169: I still don't see how the size could be negative, could you please explain in what cases this could happen?

adrianardv: It shouldn't happen, but if I'm not wrong, in some cases no data can be assigned to a cost/agent.
e.g. 
- PyTorchCost does not reject an empty dataset (n_samples=0) when batch_size="all" .
- In PyTorchDatasetHandler, with _random_split or _heterogeous_split and smaples_per_partition=0
- PyTorchDatasetHandler with random split, samples_per_partition=None, and n_partitions>n_samples
- KaggleDatasetHandler with samples_per_partition=0, or samples_per_partition=None and n_partitions>len(df)

nicola-bastianello: ok, I'll look into these. for now I'm merging with this check in place, it doesn't hurt

Issue #248, opened 2026-02-20
Implement Generic Schedulers for Learning Rates, Agent State Snapshots, and Other Parameters


Simpag:
## Summary
Create a set of generic scheduler classes/utilities for configuring and managing schedules for learning rates, agent state snapshotting intervals, and similar periodic or adjustable behaviors in the benchmarking system.

## Motivation
- Improve flexibility and modularity for experimenting with different scheduling strategies (e.g., step, exponential, cosine decay) for learning rates and other configurable parameters.
- Standardize scheduling logic so it can be reused across different agents and components (learning rates, snapshot periods, etc).
- Facilitate easy extension for future needs.

---
If further discussion is needed about the requirements, possible interfaces, or backward compatibility, please comment below.

PR #301, opened 2026-04-26
feat(examples): Examples, bug fixes and performance improvements

Simpag:
This PR includes bug fixes and performance improvements discovered during my thesis experimental phase. Refactoring and examples are also included.

Bug fixes:
* Make PyTorchCost picklable + reduce pickle size
* Add `max_batch_size` to `PyTorchCost` so that no OOM issues occur when `indices = "all"`. This happens when the dataset is large and wont fit into memory.
* Prevent `display_metrics` from crashing when trying to use `ComputationalCost` while no agent metric views are loaded.
* Prevent "prediction" metrics from crashing when PyTorch model diverges and produces inf/nan predictions.
* Dont store `agent_metric_views` with `metric_result` as this is redundant information, drastically reduces file size for large datasets.

Performance:
* Clear metric utils caches after metric computation to prevent memory leak
* Remove deepcopy from `display_metrics` as the `metric_result` objects can be large, preventing OOM issues in some cases. Same functionality is kept, the provided `metric_result` object is restored before function termination. 
* Move message compression from `_send_one` to `send` which compresses the message once and uses the same compressed message for all transmissions instead of re-computing it every time.
* Update compression schemes to be 10-25x faster depending on the scheme and message size (native level operations might improve this even more).
* Add MPS support for PyTorchCost (macbook accelerator) which can be useful for medium to large neural networks.

Quality of life:
* `AgentHistory` now accepts `-1` to get the last recorded state
* Add iop method to check if object is a valid Array object (since its unwrapped during runtime, might remove after iop refactor)
* More feedback when loading/saving large checkpoints
* Add `show_plots` argument to `display_metrics` so users can plot the metrics without showing them. This is useful when multiple benchmark executions are ran sequentially and users wants to create the plots without blocking (execution halts until plots are closed).
* CheckpointManager now compresses checkpoints with close to 0 extra overhead but massively reduced checkpoint sizes.
* New CheckpointManager methods to help with automatic decisions for resuming or starting benchmarks from scratch: `is_benchmark_started`, `is_benchmark_completed`, `are_metrics_computed`. Maybe this logic should be built into resume_benchmark function?
* `checkpoint_size` method added to CheckpointManager to get size of checkpoint.
* Add max length to algorithm names in the main benchmark progress bar
* When running more algorithms than can fit in the terminal, the progress bar will automatically scroll to prevent weird behavior.

Refactor:
* Split algorithms into multiple folders for easier management

New content:
* New algorithms such as LT-ADMM, GT-SAGA etc...
* PyTorch optimizer support for `PyTorchCost`
* Add Loss metric
* MNIST classification example using the PyTorchCost
* Neural-Implicit-Mapping example using the PyTorchCost

Misc:
* Temporary document showing how users can make use of PyTorch optimizers for local training.
* Added more tests
* Other minor changes, bug fixes, performance improvements and memory management


decent_bench/utils/interoperability/_functions.py
```diff
@@ -112,6 +112,9 @@ def to_torch(array: Array | SupportedArrayTypes, device: SupportedDevices, dtype
 
     value = array.value if isinstance(array, Array) else array
     framework_device = device_to_framework_device(device, SupportedFrameworks.PYTORCH)
+    if dtype == np.float64:
+        # PyTorch does not support float64 on some devices, so we convert to float32 if float64 is requested
+        dtype = torch.float32
```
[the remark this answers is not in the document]
Simpag: There is no point in using anything higher than float32 when working with neural networks

decent_bench/algorithms/p2p/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+__all__ = [
+    "ADMM",
+    "ATC",
+    "ATCT",
+    "ATG",
+    "DGD",
+    "DLM",
+    "ED",
+    "EXTRA",
+    "GTSAGA",
```
nicola-bastianello: I liked the use of underscores in the names you had when I looked at the branch. I would change to `GT_SAGA`, `GT_SARAH`, `GT_VR`, `LT_ADMM` `LT_ADMM_VR` `ADMM_Tracking` `ADMM_TrackingGradient` `ATC_DIGing` `ATC_Tracking`

decent_bench/algorithms/p2p/_dinno.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # Step 2: Dual variable update (line 10) - Equation (4a)
+        for i in network.active_agents():
+            self._auxiliary_update(i)
+
+        # Step 3: Approximate primal update (lines 11-15) - Equation (4b)
+        for i in network.active_agents():
+            self._local_training(i)
+
+    def _communication(self, agent: Agent, network: P2PNetwork) -> None:
```
nicola-bastianello: why introduce a private function when it's only a one-liner? wouldn't it be slightly more efficient to do it in the `step` directly?

decent_bench/algorithms/p2p/_gt_saga.py
```diff
@@ -0,0 +1,168 @@
...
+from ._p2p_algorithm import P2PAlgorithm
+
+
+@tags("peer-to-peer", "gradient-tracking")
+@dataclass(eq=False)
+class GTSAGA(P2PAlgorithm):
+    """
+    Gradient Tracking with SAGA variance reduction :footcite:p:`Alg_GT_SAGA_2020` :footcite:p:`Alg_GT_SAGA_2022`.
+
+    Warning:
```
nicola-bastianello: we should maybe document it as `Raises` given that it does so when the cost is not empirical

this raise is a bit late in terms of the benchmark pipeline, but for now it will do. maybe in teh future we could have benchmark check the tags of the algorithms, and if they have "empirical-risk" or "variance-reduction" then it raises

decent_bench/algorithms/p2p/_gt_saga.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def _update_gradient_estimator(self, agent: Agent) -> None:
+        """
+        Update local stochastic gradient estimator using SAGA variance reduction.
+
+        Raises:
+            TypeError: If the agent's cost function is not an instance of EmpiricalRiskCost.
+
+        """
+        if TYPE_CHECKING:
```
nicola-bastianello, line 115: I'm guessing mypy is to blame for this?

nicola-bastianello (commented):
thank you! this is the first batch of comments on algorithms, more will come

decent_bench/utils/progress_bar.py
```diff
@@ -195,7 +195,10 @@ def __init__(  # noqa: PLR0917
         )
         self.progress_step = progress_step
         p_cols = [
-            (TextColumn("{task.description}"), Text("Algorithm", style="bold")),
+            (
+                TextColumn("{task.description}", table_column=Column(no_wrap=True, max_width=24)),
```
nicola-bastianello, line 199: if I understand correctly this could potentially truncate algorithm names beyond 24 characters, right? what if we set no_wrap=False so that the full alg name is displayed on possibly several rows?

decent_bench/utils/pytorch_utils.py
```diff
@@ -135,3 +137,28 @@ def __init__(self, dim: int = 1):
     def forward(self, x: torch.Tensor) -> torch.Tensor:
         """Forward pass applying argmin."""
         return torch.argmin(x, dim=self.dim)
+
+
+def visualize_scheduler(scheduler: torch.optim.lr_scheduler.LRScheduler, num_iterations: int) -> None:
```
nicola-bastianello: is this used anywhere? if not, I think it would be better to omit it for now as it is not part of the core functionality and most users will not have a need for it

when we address #248 we could introduce a more general version of this visualizer

decent_bench/schemes.py
```diff
@@ -175,63 +175,91 @@ class Quantization(CompressionScheme):
...
+            scale = np.power(10.0, self.n_significant_digits - 1 - magnitudes)
+            msg_np[mask] = np.round(msg_np[mask] * scale) / scale
+
+        return iop.to_array_like(msg_np, msg)
 
 
 class TopK(CompressionScheme):
     """
-    Top-k compression which transmits only the k elements with largest magnitude.
+    Top-k compression which transmits only the top k fraction of elements by magnitude.
```
nicola-bastianello: I would change the docstring a bit to say that  k can be an int or a float, and the corresponding behaviors

docs/source/_extensions/class_tagger.py
```diff
@@ -15,7 +15,7 @@
 Add the modules to scan::
 
     tagged_list_modules = [
-        "decent_bench.distributed_algorithms",
+        "decent_bench.algorithms.p2p",
```
nicola-bastianello, line 19: federated algorithms are also tagged, so I guess this should be `"decent_bench.algorithms"`

nicola-bastianello (commented):
another batch of comments. I only need to go through the examples which I'll do tomorrow or next week

examples/nim/__init__.py
nicola-bastianello, line 1: why an init here?

examples/nim/src/dataset.py
```diff
@@ -0,0 +1,313 @@
...
+
+from decent_bench.datasets import DatasetHandler
+from decent_bench.utils.array import Array
+from decent_bench.utils.logger import LOGGER, start_logger
+from decent_bench.utils.types import Dataset
+
+
+def image_to_occupancy(
+    image_array: NDArray[np.float64],
+    threshold: float = 0.5,
```
nicola-bastianello: we have a very sharp image, so I think a threshold of 0.5 captures the occupancy very well. however, if someone uses a less sharp image (with some grey values) 0.5 might lead to some false positives (obstacles that are classified as empty space)

this is not a big issue, but we could have a more conservative default threshold e.g. 0.8

examples/nim/src/dataset.py
```diff
@@ -0,0 +1,313 @@
...
+    """
+    return (image_array < threshold).astype(np.uint8)
+
+
+class NIMDatasetHandler(DatasetHandler):
+    """
+    Neural Implicit Mapping Dataset.
+
+    Args:
+        image_file: Path to the input image file (grayscale, where black=occupied and white=free).
```
nicola-bastianello: maybe "where black above occupancy_threshold=occupied"

examples/nim/src/dataset.py
```diff
@@ -0,0 +1,313 @@
...
+        n_partitions: Number of spatial partitions to divide the data into.
```
nicola-bastianello, line 59: maybe mention that the partitions are square and how they are arranged over the space

examples/nim/src/dataset.py
```diff
@@ -0,0 +1,313 @@
...
+        samples_per_partition: Optional max number of samples per partition after balancing. If None, keep all samples.
+        transform: Optional function to apply to features (coordinates) before normalization. Should take a tuple
```
nicola-bastianello: maybe mention an example of what this transform could be? I guess rescaling could be one?

examples/nim/src/dataset.py
```diff
@@ -0,0 +1,313 @@
...
+            (x, y) and return an array.
+        label_transform: Optional function to apply to labels before returning. Should take an int label and return
+            an array.
+        label_balance: Optional ratio of majority to minority class in each partition after balancing. Attempts to
+            balance labels as closely to 1:1 as possible, with a maximum ratio of 1:label_balance. If None, no
+            balancing is done. Must be >= 1 if set.
+        occupancy_threshold: Threshold for converting grayscale values to occupancy labels. Defaults to 0.5.
+        leakage: Fraction of partition size to allow for overlapping samples between partitions.
```
nicola-bastianello: maybe call it "overlap" instead?

examples/nim/src/dataset.py
```diff
@@ -0,0 +1,313 @@
...
+            Defaults to 0 (no overlap).
+
+    """
+
+    def __init__(
```
nicola-bastianello, line 70: generally, how important were the `transform`, `label_balance` args in your experiments? if they are not strictly needed, it might be better to remove them to simplify the code a bit and make the example easier to understand

examples/nim/src/dataset.py
```diff
@@ -0,0 +1,313 @@
...
+        if self.label_balance is not None and self.label_balance < 1:
+            raise ValueError("label_balance must be >= 1")
+
+        # Load and process the image
+        image = Image.open(self.image_file).convert("L")
+        image_array = np.array(image, dtype=float64) / 255.0
+        self.height, self.width = image_array.shape
+        self.feature_norm = max(self.height, self.width)
+
+        # If configured to use provided paths for LIDAR sampling, do that and return
```
nicola-bastianello: comment is left over from previous version?

nicola-bastianello (commented):
the examples look great! just a few comments on the NIM one

my Python install died after upgrade to windows 11 so I was not able to run them yet. I will try to fix this and run them (likely next week)

decent_bench/algorithms/p2p/__init__.py
Simpag: I was debating this, it does make it more readable but it goes against pep naming convention. I suppose we can ignore it for algorithms

decent_bench/algorithms/p2p/_dinno.py
Simpag: This was implemented a very long time ago, I was auto-piloting when I converted those implementations to the new style. I'll update it

decent_bench/algorithms/p2p/_gt_saga.py
Simpag: Checking on tags is a bit hacky but would work. 

A more robust thing to do would be to do a single iteration of each algorithm before the benchmark starts. Could reuse the same network for all algorithms as we dont care about the result and we allow agents to be re-initialized

decent_bench/algorithms/p2p/_gt_saga.py
Simpag: Yes, otherwise `cost.batch_used` is not properly defined

decent_bench/utils/progress_bar.py
Simpag: I tried allowing names to span multiple rows but I thought it looked ugly and would fill up the terminal rather quickly. 24 max width is rather long and the only time you would have longer is if you are doing a parameter sweep or something similar, for example `LT-ADMM-SS1-0.01-SS2-0.1-P-0.5` would become `LT-ADMM-SS1-0.01-SS2-...`. The progress bar itself does not give much information other than knowing when a benchmark finishes so I didn't think that the algorithm name was that important, more so that you see progress is being made

decent_bench/utils/pytorch_utils.py
Simpag: This was left behind from when I played around with pytorch schedulers for learning rates. We could remove it as the optimizer+scheduler is a more advanced feature

docs/source/_extensions/class_tagger.py
Simpag: Its only the doc string but I can add it

examples/nim/__init__.py
Simpag: Because it used to be a normal python script and I forgot it, I'll remove it since its not needed anymore

examples/nim/src/dataset.py
Simpag: Its mostly meant to move from numpy to torch. Could also be used to extend or squeeze dimensions. The dataset is normalized to [0,1] coordinates based on the largest dimension (width or height) as neural networks tend to perform better in this setting.

examples/nim/src/dataset.py
Simpag: The transforms are needed to move from numpy to torch and to get the proper shapes of them. I can hardcode this into the dataset but that would not make it usable for numpy based cost functions. I dont mind doing that, what do you think?

examples/nim/src/dataset.py
Simpag: yep

Simpag:
I'm not sure why the windows tests failed. It passes on my machine both in linux and windows. Also not sure why it would interact with tkinter, maybe through some matplotlib import which causes a weird warning message? I will try to re-run it and if it fails again I will investigate further...

decent_bench/algorithms/p2p/__init__.py
nicola-bastianello: yes, I think we should ignore it since algorithm names can be quite unreadable otherwise

decent_bench/algorithms/p2p/_gt_saga.py
nicola-bastianello: yes, that could work. I'll make a note of that

decent_bench/utils/progress_bar.py
nicola-bastianello: ok, good point! let's keep no_wrap=True then

examples/nim/src/dataset.py
nicola-bastianello: thanks for clarifying. I would maybe hard-code the transformation numpy to pytorch and remove this arg to simplify understanding of the example. it's not difficult for users who want to do transformations to subclass and implement them

examples/nim/src/dataset.py
nicola-bastianello: I think this example really only makes sense for pytorch since a NN is needed, and numpy is not the best tool to train NNs. so yes, I would hard-code them

nicola-bastianello:
> I'm not sure why the windows tests failed. It passes on my machine both in linux and windows. Also not sure why it would interact with tkinter, maybe through some matplotlib import which causes a weird warning message? I will try to re-run it and if it fails again I will investigate further...

looks like it passes now! in the past I had (maybe) the same issue, and I think it was due to windows pytest timing out. I think windows still doesn't play nicely with the test suite, on my pc it is always quite slow

Simpag:
> > I'm not sure why the windows tests failed. It passes on my machine both in linux and windows. Also not sure why it would interact with tkinter, maybe through some matplotlib import which causes a weird warning message? I will try to re-run it and if it fails again I will investigate further...
> 
> looks like it passes now! in the past I had (maybe) the same issue, and I think it was due to windows pytest timing out. I think windows still doesn't play nicely with the test suite, on my pc it is always quite slow

Windows is always a low priority when it comes to performance. It takes a little less than 3 minutes to run pytest on my machine. On linux (WSL) it takes about 30 seconds so if you can, install WSL. Its rather simple

Simpag:
Forgot to mention but checkpoint manager now implicitly stores agent_metrics (the agent metric views) to reduce checkpoint size. 

Also, the regularization parameter to the PyTorchCost.local_training(..) accepts functions which takes the optimization parameters as its parameter. A simple example of L2 regularization is shown in test_pytorch_cost.py. This regularization parameter is only applied to the local_training(..) execution and not to the other methods (function, gradient etc). If users want global regularization they should include that in the loss function supplied at initialization. This should be mentioned when the user-docs are extended

PR #304, opened 2026-04-28
fix(fednova stability): skip empty FedNova normalizer rounds under impairments

### LINKED ISSUE #303: FedNova unstable under network impairments
FedNova is not able to properly handle all network impairments. I not have looked into why it fails but running the following test `pytest test/test_distributed_algorithms.py::test_fed_algorithm_execution[FedNova-kwargs3-LogisticRegressionCost-True]` will sometimes fail with error: `RuntimeError: FedNova server did not receive any normalizer uploads in this round` from `_collect_received_normalizers(...)`.

Should be looked into after PR #301 is merged

adrianardv:
Fix FedNova so that rounds where no first phase `a_i` uploads are received are treated as valid skipped rounds instead of raising a runtime error.

#### Changes:
- FedNova no longer raises when all normalizer uploads are dropped in a round.
- If phase 1 receives no `a_i `uploads:
      - the round exits early
      - cached client side cumulative gradients are cleared
      - the server model is left unchanged
 
Closes #303


nicola-bastianello:
looks good, thank you! I'll wait to merge Simon's PR and then merge this one (I can fix the conflicts myself)

Simpag:
I have split the algorithms up into different folders and extended some tests, might want to hold off on the merge until my PR is completed but the fix LGTM

adrianardv:
I tried reruning pytest windows several times, but it keeps failing. It passes on my computer

nicola-bastianello:
this is what copilot has to say about it. we could try with fix 1., as I suspect it's a timeout issue due to the slow speed of pytests on windows

-------------------------------------------------------------------------------------------------------------------------

I'll help you find a solution for the failing job. Let me start by getting the logs for that job.The logs show a critical Windows fatal exception (code 0x80000003 - which is a breakpoint exception) occurring during pytest execution. Let me examine the workflow configuration and test files to better understand the issue.## Solution for Job 73652171507

**Issue:** The pytest job on Windows is crashing with a fatal exception (exit code 2147483651) during test execution. The error occurs after approximately 161 seconds of running tests, suggesting either a resource exhaustion issue, a flaky test that's timing out, or an environment-specific problem on Windows.

**Root Cause Analysis:**

The Windows fatal exception `0x80000003` combined with the exit code indicates a process crash, likely due to:
1. **Memory exhaustion** - Tests consuming too much memory on Windows runners
2. **Timeout or resource limits** - Windows environment hitting constraints
3. **Platform-specific test issue** - A test behaving differently on Windows

The pytest configuration in `pyproject.toml` (lines 100-103) has minimal settings:
```toml
[tool.tox.env.pytest]
description = "Run pytest (test executor)"
deps = ["pytest"]
commands = [["pytest"]]
```

**Recommended Solutions:**

1. **Add pytest timeout configuration** to prevent hanging tests and catch resource issues earlier:
```toml
[tool.pytest.ini_options]
addopts = "-q -W error --timeout=300"
testpaths = ["test"]
```
   (Requires adding `pytest-timeout` as a dev dependency)

2. **Run pytest with verbose output** to identify which test fails:
```yaml
- name: Run pytest
  run: tox -e pytest -- -v
```

3. **Add memory limits or parallel execution control** for Windows by updating pyproject.toml:
```toml
[tool.tox.env.pytest]
description = "Run pytest (test executor)"
deps = ["pytest"]
commands = [["pytest", "-n", "auto"]]
```
   (Requires `pytest-xdist` for parallel execution, which can help with resource issues)

4. **Check for platform-specific test failures** - Review tests that may be sensitive to Windows file systems or memory management.

**Immediate Action:** Update the pytest tox environment to include verbose output and timeout protection to get better diagnostics on the next run.

Simpag:
The test that fails is `test/test_distributed_algorithms.py::test_p2p_algorithm_execution[WangElia-kwargs5-LogisticRegressionCost-True]` with some complaints about tkinter. Tkinter is only used by matplotlib in this project if I'm not mistaken. This is really weird because that test just creates a bare network and runs alg.run() which shouldn't interact with tkinter. One thing it could be is if we have tests testing `display_metrics` which opens plots that are never closed. Then when gc kicks in it tries to close the plots causing the crash. Double check if you can find any such tests and make sure to close all matplotlib figures after the call to display_metrics

adrianardv:
What I don't understand is why it is failing here and not in the other 2 open PRs. The failing test doesn't seem related to the changes introduced in this PR

Simpag:
Because we love windows :)

Simpag:
I reran the test with the extended (not really but a little) GitHub actions debug logs and it seems like it has something to do with the centralised solver. I am not currently at home so I can't really investigate much further right now. Maybe you can find something @adrianardv ?

```shell
File "D:\a\decent-bench\decent-bench\decent_bench\costs\_empirical_risk\_logistic_regression_cost.py", line 255 in gradient
File "D:\a\decent-bench\decent-bench\decent_bench\utils\interoperability\_decorators.py", line 99 in wrapper
File "D:\a\decent-bench\decent-bench\decent_bench\centralized_algorithms.py", line 244 in step
File "D:\a\decent-bench\decent-bench\decent_bench\centralized_algorithms.py", line 157 in run
File "D:\a\decent-bench\decent-bench\decent_bench\centralized_algorithms.py", line 60 in solve
File "D:\a\decent-bench\decent-bench\decent_bench\benchmark\_utils.py", line 109 in create_classification_problem
```

adrianardv:
All checks are passing now!

1. The original Windows failure was Tk/matplotlib unraisable exceptions during later distributed tests. So, in `test_display.py`, I made sure matplotlib figures are closed after the display tests. I also adjusted a couple of helper-level plotting tests so they do not open GUI-backed pyplot figures. This should not affect normal plotting behavior, because the change is only in tests and does not change the library plotting backend.
2. Then, the crash happenned while Rich progress bar was active during `create_classification_problem(...) -> ca.solve(...)` inside `test_distributed_algorithms.py`. Since those tests are checking algorithm execution, not terminal progress output, I added a `show_progress` option to `solve(...)` and the problem builders. It defaults to `True`, so normal user behavior should stay the same. In `test_distributed_algorithms.py` only, I pass `show_progress=False`.

The part I am less certain about is the progress-bar change. It seems reasonable because the distributed tests do not need progress output, and there is now a small test that still checks `show_progress=True` works. But this is more of a CI stability fix than a full investigation into Rich progress behavior on Windows.

> What I don't understand is why it is failing here and not in the other 2 open PRs. The failing test doesn't seem related to the changes introduced in this PR

About this, copilot says "this is timing/order/environment dependent: Rich progress rendering and Tk object cleanup involve background threads / garbage collection, so whether the crash appears can depend on runner speed, test order, or small branch differences even when the failing code path is mostly shared."

Simpag:
Not really sure what is going on but glad it works now. I don't think its worth spending too much time investigating it, windows + pytest environment is not the best of combo it seem

PR #305, opened 2026-04-29
feat(distributed_algorithms): Implement Fed-LT federated algorithm

adrianardv:
### Summary

- Add `FedLT` as a federated algorithm with local-training updates, auxiliary `z_i` variables, partial participation support, and optional accelerated local updates.
- Implement the Fed-LT server update using the server cost proximal operator for the optional global regularizer term `prox_{rho h / N}`.
- For partial participation or dropped uploads, the server keeps the last stored `z_i` for clients whose fresh `z_i` was not received, matching the stale-value aggregation interpretation.
- GD vs GD is cost-driven rather than exposed as separate solver modes: `EmpiricalRiskCost.gradient(...)` uses its default mini-batch behavior, while generic `Cost.gradient(...)` performs full-gradient updates.
- Acceleration is exposed through `use_acceleration=True` and requires finite non-negative `m_smooth` and `m_cvx` metadata on every client cost, with `m_smooth >= m_cvx`.
- Add focused FedLT tests covering initialization, parameter validation, local updates, acceleration, partial participation, empirical-risk batching, server proximal handling, and noise/compression smoke behavior.


decent_bench/algorithms/federated/_fed_lt.py
```diff
@@ -0,0 +1,214 @@
...
+    After local training, the client sets
+
+    .. math::
+        x_{i,k+1}=w^{N_e}_{i,k}, \qquad z_{i,k+1}=z_{i,k}+2(x_{i,k+1}-y_{k+1}),
+
+    and uploads :math:`z_{i,k+1}` to the server. Inactive clients keep
+    :math:`x_{i,k+1}=x_{i,k}` and :math:`z_{i,k+1}=z_{i,k}`. For later server averages, the server stores a received
+    fresh :math:`z_{i,k+1}` when the upload arrives and otherwise keeps its previous stored :math:`z_i`.
+
+    Fed-PLT is the privacy-noise version of Fed-LT :footcite:p:`Alg_FedPLT`. This class does not add Gaussian noise
```
nicola-bastianello: I would remove everything after `Fed-PLT is the privacy-noise version of Fed-LT :footcite:p:`Alg_FedPLT`.` as is might be a bit confusing to think of communication noise as equivalent to the privacy mechanism. in my view privacy mechanisms are to be embedded in the algorithm, while communication noise is something due e.g. wireless interference. and compression doesn't have to be mentioned explicitly here

decent_bench/algorithms/federated/_fed_lt.py
```diff
@@ -0,0 +1,214 @@
...
+        if self.step_size <= 0:
+            raise ValueError("`step_size` must be positive")
+        if self.num_local_epochs <= 0:
+            raise ValueError("`num_local_epochs` must be positive")
+        if self.rho <= 0:
+            raise ValueError("`rho` must be positive")
+
+    def initialize(self, network: FedNetwork) -> None:
+        self.x0 = alg_helpers.initial_states(self.x0, network)
+        if self.use_acceleration:
```
nicola-bastianello, line 115: why not do `_validate_accelerated_constants` in post init?

decent_bench/algorithms/federated/_fed_lt.py
```diff
@@ -0,0 +1,214 @@
...
+            self._validate_accelerated_constants(network)
+
+        server = network.server()
+        z_by_client = {client: iop.copy(self.x0[client]) for client in network.clients()}
+        server.initialize(x=self.x0[server], aux_vars={"z_by_client": z_by_client})
+        for client in network.clients():
+            client_x0 = self.x0[client]
```
nicola-bastianello: z_i,0 in my paper can be chosen arbitrarily; I would add a z0 arg to the init to allow setting it. you can look at ADMM in p2p for an example of this

decent_bench/algorithms/federated/_fed_lt.py
```diff
@@ -0,0 +1,214 @@
...
+        if not participating_clients:
+            return
+        self._clear_buffered_server_messages(network, participating_clients)
+        self._run_local_updates(network, participating_clients)
+        self.aggregate(network, participating_clients)
+
+    def _compute_server_y(self, network: FedNetwork) -> "Array":
+        z_values = list(network.server().aux_vars["z_by_client"].values())
+        average_z = iop.mean(iop.stack(z_values, dim=0), dim=0)
+        return network.server().cost.proximal(average_z, self.rho / len(network.clients()))
```
nicola-bastianello, line 168: can you please update the docs of ZeroCost.proximal to say that it always returns `x`? this way the behavior is also clear from there

decent_bench/algorithms/federated/_fed_lt.py
```diff
@@ -0,0 +1,214 @@
...
+        else:
+            for _ in range(self.num_local_epochs):
+                grad = client.cost.gradient(local_x) + ((local_x - v) / self.rho)
+                local_x -= self.step_size * grad
+
+        z_next = z + (2 * (local_x - y))
+        return local_x, z_next
+
+    def _compute_accelerated_local_update(self, client: "Agent", local_x: "Array", v: "Array") -> "Array":
+        smoothness = client.cost.m_smooth + (1 / self.rho)
```
nicola-bastianello: all these values will be the same throughout the algorithm's run. since these require the computation of square roots, I suggest computing momentum during initialize so that the operation is performed only once

nicola-bastianello (commented):
great work, thank you! just some comments

decent_bench/algorithms/federated/_fed_lt.py
adrianardv: Because `_validate_accelerated_constants`needs the concrete `FedNetwork`, and at `__post_init__` time we still don't have access to the network

Issue #302, opened 2026-04-27
Support structured network messages


adrianardv:
Right now we assume that `msg`is an `Array` in many places across the network layer and distributed algorithms. This makes it awkward to implement algorithms that need to transmit multiple quantities together, such as FedNova uploads. 
Also because in the current network design, each receiver stores only one message per sender, so consecutively sending multiple messages makes later sends overwrite earlier ones.

A future enhancement would be to support structured messages, instead of forcing algorithms to:
- flatten and concatenate multiple quantities into one array, or
- perform sevaral separate transmissions

----
At the moment:
- `Network.send(...)` is designed around array messages
- many algorithms assume received messages support array operations directly

---
Possible options:
A) Support `Array | dict[str, Array]`
B) Make all messages strcutred (`dict[str, Array]`): make all messages to be dicts of arrays, even when only one quantity is sent.

For structured messages, compression/noise are applied field by field.

nicola-bastianello:
just some thoughts: if we allow messages to be both Array and dict, we might have issues with algorithms assuming that they are Array only (and I think you encountered mypy issues as well). to solve that, forcing all messages to be dict could work, but would lead to messy syntax especially since most algorithms only transmit one quantity. so I would say none of them is a perfect solution

some ways to solve it could be

- implement the iop backend so that it takes dicts applying the operation to each value; for unary operations this is perfectly doable, but for binary operations we would need to check that they are "broadcastable" (e.g. two dicts with all the same keys, or one dict and one Array). I think it is possible but it is definitely more complicated, and I don't think we should try to solve this problem in the iop layer
- so, an alternative I see is to always use dicts in the network, and when an Array is passed the network internally converts it to a dict with e.g. key "msg". so in a sense we wrap Array messages in a dict and then all operations in the network backend can be performed on dicts, and then we unwrap when placing the Array message in the recipient memory. I'm not sure if this has a performance edge on just handling Array and dict in two separate branches

one thing I would like to avoid is to type things with custom types. I don't know if it is possible (I think in your implementation draft it wasn't) but worth spending some time on

Simpag:
[...]

PR #306, opened 2026-05-04
feat(schemes): Add client selection and compression schemes

adrianardv:
- Add new federated client selection schemes for data-size, fairness, staleness,
  high-loss, and hybrid selection policies.
- Add QSGD and error-feedback compression schemes.
- Add trace-based and cyclic activation schemes.
- Add tests covering the new schemes behaviour.
- Update the user guide with clearer network constraint, compression, noise, and
  error-feedback documentation.
- Validate error-feedback compression with a local experiment comparing no
  compression, TopK compression, and error-feedback TopK. The run showed the
  expected ordering: no compression achieved the lowest final loss, plain TopK
  degraded convergence the most, and error-feedback TopK partially recovered the loss introduced by TopK compression.


decent_bench/utils/__init__.py
```diff
@@ -0,0 +1,5 @@
+from decent_bench.utils.agent_utils import infer_client_data_size
+
+__all__ = [
+    "infer_client_data_size",
+]
```
Simpag: This might shadow the other utils in the folder. Either leave the file blank or add the other utils here

decent_bench/schemes.py
```diff
@@ -171,8 +749,57 @@ def compress(self, msg: Array) -> Array:  # noqa: D102
...
+    """
+
+    def __init__(self, compression: CompressionScheme):
+        self.compression = compression
+        self._residual: Array | None = None
+
+    def compress(self, msg: Array) -> Array:  # noqa: D102
+        corrected_msg = iop.copy(msg)
+        if self._residual is not None:
+            corrected_msg += self._residual
```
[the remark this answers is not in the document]
Simpag: Maybe make residual a dict[shape, Array] cache?

decent_bench/schemes.py
```diff
@@ -171,8 +749,57 @@ def compress(self, msg: Array) -> Array:  # noqa: D102
...
+
+        compressed_msg = self.compression.compress(iop.copy(corrected_msg))
+        self._residual = corrected_msg - compressed_msg
+        return compressed_msg
+
```
Simpag: Feel free to not copy messages in schemes (also line 780) as the network takes care of copying the message. This was done for two reasons:
1. Reduces the risk of users mutating the message when we don't want to in their own schemes, they don't have to worry about it now
2. Performance, doing the copy in the network allows us to copy the message one time instead of doing it in every scheme

Since you're calculating a residual it gets a bit tricky. You might be able to get away with something like: `self._residual = corrected_msg - (compressed_msg := self.compression.compress(corrected_msg))` but I am not sure, you might need to do a copy of the corrected_msg here

decent_bench/schemes.py
```diff
@@ -192,6 +819,66 @@ def compress(self, msg: Array) -> Array:  # noqa: D102
...
+            raise ValueError("`n_levels` must be a positive integer")
+        self.n_levels = n_levels
+
+    def compress(self, msg: Array) -> Array:  # noqa: D102
+        msg_norm = float(iop.norm(msg))
+        if msg_norm == 0:
+            return iop.zeros_like(msg)
+
+        magnitudes = iop.to_numpy(iop.absolute(msg), dtype=np.float64)
+        signs = iop.to_numpy(iop.sign(msg), dtype=np.float64)
```
Simpag: would probably be more efficient to do `msg_np = iop.to_numpy(msg)` and then use np methods instead of doing two `iop.to_numpy(...)`. While I believe most frameworks share the memory when converting to numpy if the origin resides on the cpu (then its rather cheap to convert) but when you use an accelerator the conversion to numpy is costly due to the cross-device communications

Simpag (commented):
Just some minor performance suggestions

decent_bench/utils/__init__.py
adrianardv: I'm leaving it blank then

decent_bench/schemes.py
adrianardv: Good point. I'll then remove the first copy.
But I'm keeping the second one because EFC needs the original
`corrected_msg` to compute the residual, and some compressors, like `Quantization`, can mutate NumPy-backed inputs through `iop.to_numpy`, so passing `corrected_msg` directly could make the residual wrong.

decent_bench/schemes.py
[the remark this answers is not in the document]
adrianardv: Good idea! I've implemented the shape-keyed residual cache.

It is still not a perfect separation, because two different logical message streams could have the same shape. Once #302 is added, EFC should key residuals by channel/labels instead of only by shape.

decent_bench/schemes.py
```diff
@@ -124,38 +257,483 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
+            selected_clients = clients_list
+        else:
+            staleness_weight = 1.0 - self.loss_weight
+            loss_scores = [0.0 for _ in clients_list]
+            staleness_scores = [0.0 for _ in clients_list]
+            if self.loss_weight > 0:
+                loss_scores = self._normalized_scores([client.cost.function(client.x) for client in clients_list])
+            if staleness_weight > 0:
+                staleness_scores = self._normalized_scores(
+                    [self._staleness(client, iteration) for client in clients_list],
```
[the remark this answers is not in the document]
adrianardv: This is now fixed

decent_bench/schemes.py
```diff
@@ -124,38 +257,483 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
+        k = self._num_selected_clients(clients, self.clients_per_round, self.client_fraction)
+        clients_list = list(clients)
+        if k >= len(clients_list):
+            selected_clients = clients_list
+        else:
+            count_weight = 1.0 - self.loss_weight
+            loss_scores = [0.0 for _ in clients_list]
+            count_scores = [0.0 for _ in clients_list]
+            if self.loss_weight > 0:
+                loss_scores = self._normalized_scores([client.cost.function(client.x) for client in clients_list])
```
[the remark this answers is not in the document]
adrianardv: fixed

decent_bench/schemes.py
```diff
@@ -124,38 +257,483 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
+        k = self._num_selected_clients(clients, self.clients_per_round, self.client_fraction)
+        clients_list = list(clients)
+        if k >= len(clients_list):
+            return clients_list
+
+        data_size_weight = 1.0 - self.loss_weight
+        loss_scores = [0.0 for _ in clients_list]
+        data_size_scores = [0.0 for _ in clients_list]
+        if self.loss_weight > 0:
+            loss_scores = self._normalized_scores([client.cost.function(client.x) for client in clients_list])
```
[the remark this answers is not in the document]
adrianardv: fixed

decent_bench/utils/agent_utils.py
```diff
@@ -0,0 +1,62 @@
...
+    """
+    if data_size_key in client.data:
+        return _validate_positive_client_data_size(client.data[data_size_key])
+
+    cost = client.cost
+    if hasattr(cost, "n_samples"):
+        n_samples = cost.n_samples
+        if n_samples is not None:
+            return _validate_positive_client_data_size(n_samples)
+    if hasattr(cost, "A"):
```
nicola-bastianello: I've been thinking about this, and I believe that the only meaningful size we can infer is n_samples of an empirical risk cost. if we apply this to Quadratic cost, it will return the size of x, but that's not meaningful. since this method is only used when empirical risk costs are involved, I would only use n_samples

decent_bench/utils/agent_utils.py
```diff
@@ -0,0 +1,62 @@
+from typing import TYPE_CHECKING, Any
+
+import decent_bench.utils.interoperability as iop
+
+if TYPE_CHECKING:
+    from decent_bench.agents import Agent
+
+
+def _validate_positive_client_data_size(size: Any) -> float:  # noqa: ANN401
```
nicola-bastianello: I'm still not sure what this is needed for; in any case, if we change the infer util to use n_samples, this can maybe be removed

decent_bench/networks.py
```diff
@@ -140,6 +140,14 @@ def _initialize_message_schemes(
         Returns:
             dict[Agent, scheme]: mapping each agent in `self.graph` to its scheme instance.
 
+        Note:
```
nicola-bastianello: I think it would be better to not put this burden on the user, and to fix the issue in the backend; I'm guessing that assigning a deepcopy of the scheme to each node will be enough to ensure stateful schemes don't have issues

docs/source/user.rst
```diff
@@ -159,6 +159,59 @@ weights.
 :class:`~decent_bench.algorithms.federated.Scaffold` matches the standard
 SCAFFOLD algorithm and always uses uniform averaging over the selected clients.
 
+Client selection
```
nicola-bastianello, line 193: the names of these schemes are quite long; I would shorten them by only appending `Selection` instead of `ClientSelection`, I think it's clear enough

docs/source/user.rst
```diff
@@ -442,6 +495,117 @@ Create a custom benchmark problem using existing resources.
...
+    network = FedNetwork(
+        clients=clients,
+        message_compression=message_compression,
+    )
+    problem = BenchmarkProblem(
+        network=network,
+        x_optimal=x_optimal,
+    )
+
+The current compression API preserves message shape; it simulates the numerical
```
nicola-bastianello: I would remove this sentence, it's confusing for users

decent_bench/schemes.py
```diff
@@ -108,11 +129,130 @@ def is_active(self, iteration: int) -> bool:  # noqa: D102, ARG002
         return False
 
 
+class TraceActivation(AgentActivationScheme):
```
nicola-bastianello: I would not include this in the codebase. I prefer to only have schemes that work for any number of iterations, while this is iteration-dependent

of course you can use it in your thesis, and after the user guide refactor we could consider adding it there as an example

decent_bench/schemes.py
```diff
@@ -108,11 +129,130 @@ def is_active(self, iteration: int) -> bool:  # noqa: D102, ARG002
...
+    """
+    Scheme where an agent cycles through active and inactive intervals.
+
+    The agent is active for ``active_for`` iterations and inactive for ``inactive_for`` iterations in each cycle.
+    The first interval is active by default, but ``start_active=False`` starts each cycle with the inactive interval.
+
+    Args:
+        active_for: number of active iterations in each cycle.
+        inactive_for: number of inactive iterations in each cycle.
+        start_active: whether each cycle starts with the active interval.
```
nicola-bastianello: I don't see what situation would be modeled by this activation scheme; can you make an example and reference any paper? if not, maybe we should remove it

other comments:

- I'm not sure I see a situation where the start_active option will be needed; for simplicity I would remove it.
- I would allow passing only active_for, with inactive_for defaulting to the same value

decent_bench/schemes.py
```diff
@@ -108,11 +129,130 @@ def is_active(self, iteration: int) -> bool:  # noqa: D102, ARG002
...
+        return phase >= self.inactive_for
+
+
 class ClientSelectionScheme(ABC):
-    """Scheme defining how to select a subset of available clients."""
+    """
+    Scheme defining how to select a subset of available clients.
+
+    Federated algorithms call :meth:`select` once per round with the currently active clients. Implementations may be
+    stateful, but they should not mutate the client sequence they receive.
```
nicola-bastianello: > but they should not mutate the client sequence they receive

I'm guessing you encountered some issues due to this? can you give me more details please?

decent_bench/schemes.py
```diff
@@ -108,11 +129,130 @@ def is_active(self, iteration: int) -> bool:  # noqa: D102, ARG002
...
+    """
+
+    @staticmethod
+    def _validate_selection_size(
+        clients_per_round: int | None,
```
nicola-bastianello: I find the names `clients_per_round` and `client_fraction` not immediately clear; maybe we could call them `num_selected_clients` and `fraction_selected_clients` or something like that?

decent_bench/schemes.py
```diff
@@ -108,11 +129,130 @@ def is_active(self, iteration: int) -> bool:  # noqa: D102, ARG002
...
-    """Scheme defining how to select a subset of available clients."""
+    """
+    Scheme defining how to select a subset of available clients.
+
+    Federated algorithms call :meth:`select` once per round with the currently active clients. Implementations may be
+    stateful, but they should not mutate the client sequence they receive.
+    """
+
+    @staticmethod
+    def _validate_selection_size(
```
nicola-bastianello, line 180: methods `_validate_selection_size`, `_num_selected_clients`, `_normalized_scores`, `_client_loss` require documentation to ensure we can maintain them in the future

decent_bench/schemes.py
```diff
@@ -124,38 +264,479 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
+        del iteration
         if not clients:
             return []
-        if self.clients_per_round is not None:
-            k = min(self.clients_per_round, len(clients))
-        else:
-            k = max(1, int(self.client_fraction * len(clients)))  # type: ignore[operator]
-            k = min(k, len(clients))
+        k = self._num_selected_clients(clients, self.clients_per_round, self.client_fraction)
         if k >= len(clients):
```
nicola-bastianello: if I'm not mistaken, the implementation of `_num_selected_clients` already ensures that `k == len(clients)` if the number of selected clients would exceed `len(clients)`. the check can then use `==`

decent_bench/schemes.py
```diff
@@ -124,38 +264,479 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
             return list(clients)
         return random.sample(list(clients), k)
 
 
+class DataSizeClientSelection(ClientSelectionScheme):
+    """
+    Data-size weighted client selection :footcite:p:`Scheme_FedSampling`.
+
+    The scheme samples clients without replacement with probability proportional to each client's local data size.
```
nicola-bastianello, line 289: it would be nice to add the equation defining the weights, so that the docs are more self-contained

decent_bench/schemes.py
```diff
@@ -124,38 +264,479 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
+
+    Args:
+        clients_per_round: number of clients to sample per round.
+        client_fraction: fraction of provided clients to sample per round.
+        data_size_key: key used to read explicit data sizes from ``Agent.data``.
```
nicola-bastianello: if I understand correctly, allowing for the use of `data_size_key` requires the users to explicitly add `Agent.data[data_size_key]`, which I think is too much to ask. in my opinion, we should always infer the sizes in the backend, without user interaction (besides of course choosing to use an empirical risk cost)

I would remove this arg, and only infer from EmpiricalRiskCost.n_samples

decent_bench/schemes.py
```diff
@@ -124,38 +264,479 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
+        self.clients_per_round = clients_per_round
+        self.client_fraction = client_fraction
+        self.data_size_key = data_size_key
+
+    def select(  # noqa: D102
+        self,
+        clients: Sequence[Agent],
+        iteration: int,
+    ) -> list[Agent]:
+        del iteration
```
nicola-bastianello: why this? maybe you could replace `iteration: int` with `_: int` if iteration is not needed

decent_bench/schemes.py
```diff
@@ -124,38 +264,479 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
+        if not clients:
+            return []
+        k = self._num_selected_clients(clients, self.clients_per_round, self.client_fraction)
+        clients_list = list(clients)
```
nicola-bastianello: this conversion to list is not needed for the `k >= len(clients_list)` check, so it can probably be done later. also, by my previous comment `==` is probably all that's needed

decent_bench/schemes.py
```diff
@@ -124,38 +264,479 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
+        data_sizes = np.array(
+            [infer_client_data_size(client, self.data_size_key) for client in clients_list],
+            dtype=np.float64,
+        )
+        probabilities = data_sizes / data_sizes.sum()
+        selected_indices = iop.rng_numpy().choice(len(clients_list), size=k, replace=False, p=probabilities)
+        return [clients_list[int(index)] for index in selected_indices]
+
+
+class ParticipationFairClientSelection(ClientSelectionScheme):
```
nicola-bastianello: I would simplify the name to `FairSelection`, ParticipationFair is not quite clear

decent_bench/schemes.py
```diff
@@ -124,38 +264,479 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
+    """
+    Participation-fair client selection :footcite:p:`Scheme_FairFedCS`.
+
+    The scheme prioritizes clients with fewer past selections, breaking ties at random.
```
nicola-bastianello: here also it would be good to show the math

decent_bench/schemes.py
```diff
@@ -124,38 +264,479 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
+    ) -> list[Agent]:
+        del iteration
+        if not clients:
+            return []
+        k = self._num_selected_clients(clients, self.clients_per_round, self.client_fraction)
+        clients_list = list(clients)
+        if k >= len(clients_list):
+            selected_clients = clients_list
+        else:
+            tie_breakers = iop.rng_numpy().permutation(len(clients_list))
```
nicola-bastianello: what if we just `sorted`, without explicitly breaking ties? after all, the sort order for clients with the same count will be essentially random from the perspective of the user

decent_bench/schemes.py
```diff
@@ -124,38 +264,479 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
+                key=lambda index: (self._selection_counts.get(clients_list[index], 0), int(tie_breakers[index])),
+            )
+            selected_clients = [clients_list[index] for index in ranked_indices[:k]]
+
+        for client in selected_clients:
+            self._selection_counts[client] = self._selection_counts.get(client, 0) + 1
+        return selected_clients
+
+
+class StaleClientSelection(ClientSelectionScheme):
```
nicola-bastianello: the difference between this scheme and `ParticipationFairClientSelection` seems to be quite small, and it looks like they are essentially equivalent. I would only keep `ParticipationFairClientSelection`

decent_bench/schemes.py
```diff
@@ -124,38 +264,479 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
+            dtype=np.float64,
+        )
+        probabilities = data_sizes / data_sizes.sum()
+        selected_indices = iop.rng_numpy().choice(len(clients_list), size=k, replace=False, p=probabilities)
+        return [clients_list[int(index)] for index in selected_indices]
+
+
+class ParticipationFairClientSelection(ClientSelectionScheme):
+    """
+    Participation-fair client selection :footcite:p:`Scheme_FairFedCS`.
```
nicola-bastianello: could you point me to where this scheme is defined in the reference? with a quick read I wasn't able to find it

decent_bench/schemes.py
```diff
@@ -124,38 +264,479 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
+        for client in selected_clients:
+            self._last_selected_iterations[client] = iteration
+        return selected_clients
+
+
+class HighLossClientSelection(ClientSelectionScheme):
+    """
+    High-loss client selection :footcite:p:`Scheme_PowerOfChoice`.
+
+    The scheme evaluates each client's local loss at its current ``x`` and selects the clients with highest loss,
```
nicola-bastianello: if I'm not mistaken, this scheme differs slightly from the one in the paper, since `_client_loss` uses client.x to evaluate the loss, while in the paper the server sends the updated model, and the clients evaluate the loss there. this is fine with me, since otherwise we would need to do communication from inside the scheme and that's not a good idea. but I would mention that the scheme is "inspired" by the paper, not exactly the same.

I would also show the math

decent_bench/schemes.py
```diff
@@ -124,38 +264,479 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
+        losses = [self._client_loss(client) for client in clients_list]
+        tie_breakers = iop.rng_numpy().permutation(len(clients_list))
+        ranked_indices = sorted(
+            range(len(clients_list)),
+            key=lambda index: (-losses[index], int(tie_breakers[index])),
+        )
+        return [clients_list[index] for index in ranked_indices[:n_selected_clients]]
+
+
+class HybridFairHighLossClientSelection(ClientSelectionScheme):
```
nicola-bastianello: I would not include this in the codebase. you can of course use it in the thesis, and we could show it in the user guide as an example later on

decent_bench/schemes.py
```diff
@@ -124,38 +264,479 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
+                key=lambda index: (-scores[index], int(tie_breakers[index])),
+            )
+            selected_clients = [clients_list[index] for index in ranked_indices[:k]]
+
+        for client in selected_clients:
+            self._last_selected_iterations[client] = iteration
+        return selected_clients
+
+
+class CountFairHighLossClientSelection(ClientSelectionScheme):
```
nicola-bastianello: I would not include this in the codebase. you can of course use it in the thesis, and we could show it in the user guide as an example later on

decent_bench/schemes.py
```diff
@@ -124,38 +264,479 @@ def select(self, clients: Sequence["Agent"], iteration: int) -> list["Agent"]:
...
+                key=lambda index: (-scores[index], int(tie_breakers[index])),
+            )
+            selected_clients = [clients_list[index] for index in ranked_indices[:k]]
+
+        for client in selected_clients:
+            self._selection_counts[client] = self._selection_counts.get(client, 0) + 1
+        return selected_clients
+
+
+class DataSizeHighLossClientSelection(ClientSelectionScheme):
```
nicola-bastianello: I would not include this in the codebase. you can of course use it in the thesis, and we could show it in the user guide as an example later on

decent_bench/schemes.py
```diff
@@ -171,8 +752,58 @@ def compress(self, msg: Array) -> Array:  # noqa: D102
         return msg
 
 
+class ErrorFeedbackCompression(CompressionScheme):
```
nicola-bastianello: I would not include this in the codebase. you can of course use it in the thesis, and we could show it in the user guide as an example later on

my concern is that error feedback belongs to algorithm design, not as a compression scheme. and different algorithms implement error feedback in different ways, with different theoretical guarantees

decent_bench/schemes.py
```diff
@@ -192,6 +823,67 @@ def compress(self, msg: Array) -> Array:  # noqa: D102
         return iop.to_array_like(msg_np, msg)
 
 
+class QSGDCompression(CompressionScheme):
```
nicola-bastianello: I would change the name, since QSGD refers to the whole algorithm proposed in that paper. a possible alternative name could be `StochasticQuantization`

I would also expand a bit the discussion of what `n_levels` is (e.g. mention if higher then .. if lower then ..)

nicola-bastianello: we should also add a warning that this scheme computes the norm, and hence might be computationally expensive

nicola-bastianello (commented):
great work, thank you for adding so many schemes! I have a lot of comments, but most of them are minor

decent_bench/schemes.py
Simpag: This scheme is not iteration-dependant. The trace list is indexed cyclicly

decent_bench/schemes.py
Simpag: Could possibly be used to simulate charge cycles on robots

decent_bench/schemes.py
nicola-bastianello: ah yes, with the repeat option it repeates cyclically. I still think we should remove it as it might be confusing and it's nonstandard

decent_bench/schemes.py
nicola-bastianello: good point

decent_bench/utils/agent_utils.py
adrianardv: Okey. I removed the `A`/`b` fallbacks.

decent_bench/utils/agent_utils.py
adrianardv: I removed the separate validation helper. But I kept inline validation for the explicit `Agent.data[...]` path. As the value is provided by the user it could be invalid.

Edit: as per your later comment I removed the `Agent.data[...]` path and the validation helper

decent_bench/schemes.py
adrianardv: I had also thought about mobile devices that become available during recurring charging/idle windows.

I remember having read this when I did the literature review, and I also found these papers now: 
- "On the Convergence of Federated Averaging
with Cyclic Client Participation" https://arxiv.org/pdf/2302.03109
- "Federated Learning under Periodic Client Participation and Heterogeneous Data: A New Communication-Efficient Algorithm and Analysis" https://arxiv.org/abs/2410.23131

I made inactive_for optional, defaulting to active_for, as you suggested.
Thinking about start_active, I removed it; and I added an offset parameter. It seems better to model staggered cycles.
For example, different phones or robots can follow the same active/inactive cycle but start at different phases.

decent_bench/schemes.py
adrianardv: I added it as a precaution. If I'm not mistaken, there is a possible accidental side effect.
A selector should only choose and return clients, but if it sorted/shuffled/popped from the input list in-place, it could modify the active client list (in the network).

To make this safer, I could also pass `list(network.active_clients())` to `select()` instead of pasing `network.active_clients()` directly.

decent_bench/schemes.py
adrianardv: Okey, I think that is reasonable here. Since this scheme only balances participation counts, sorting by count can be enough. Once a client is selected, its count increases and the next equal-count clients are preferred in later rounds.

decent_bench/schemes.py
adrianardv: `StaleSelection` tracks time since last selection, while `FairSelection` tracks total number of selections, so they are not exactly identical, but, yes, the difference is small. I'll take it out.

decent_bench/schemes.py
adrianardv: FairFedCS is more complex, the FairSelection I implemented is a simplification inspired by FairFedCS.

In Section 3.2., around equation 3 and 5, their queue dynamic increases if the client is not selected to join a round. Then, they use that queue in the index used for client selection (in eq. 13).

I'll adjust the wording so the citation is framed as motivation.

decent_bench/schemes.py
adrianardv: Yes.
I updated the docstring to say "inspired by", and added the selection equation.

decent_bench/schemes.py
adrianardv: I see the algorithm dependent issue.
My motivation was that EF is a common wrapper around compressors, so it seemed useful to expose it as a plug in scheme.
If I'm not wrong, without this scheme, an algorithm that needs EF has to keep the residual in the algorithm, compress the residual-corrected update manually, and send the message through a network configured with no compression, right?

decent_bench/schemes.py
nicola-bastianello: I think it's ok to keep it this way

decent_bench/schemes.py
nicola-bastianello: yes, that's correct. compression and error feedback would both exist in the algorithm. it would be nice to use the network compression scheme, but as I said a generic implementation of error feedback might not enough in many cases

but I do see the benefit in using the netwokr compression and having some way to easily plug in error feedback. let's discuss later how we could do this (but, I think, acting on the Algorithm object)

nicola-bastianello:
looks great, thank you! merging now

PR #308, opened 2026-05-08
feat(federated): Add FedPD, FedDyn, and update message buffering

adrianardv:
- Add FedPD and FedDyn algorithms
- Extend FedLT with an Adam local solver
- Allow Scaffold control variate initialization

- Remove 'Network(buffer_messages=...)`. Received messages are always cleared at network step boundaries.

Closes #307

### DIFF
--- a/decent_bench/algorithms/federated/_fed_algorithm.py
+++ b/decent_bench/algorithms/federated/_fed_algorithm.py
@@ -74,41 +74,12 @@ class FedAlgorithm(Algorithm[FedNetwork]):
         return self.selection_scheme.select(active_clients, iteration)
 
     def server_broadcast(self, network: FedNetwork, selected_clients: Sequence["Agent"]) -> None:
-        """Clear stale server broadcasts, then send the current server model to the selected clients."""
-        self._clear_buffered_client_server_messages(network, selected_clients)
+        """Send the current server model to the selected clients."""
         network.send(sender=network.server(), receiver=selected_clients, msg=network.server().x)
 
     def _clients_with_server_broadcast(self, network: FedNetwork, selected_clients: Sequence["Agent"]) -> list["Agent"]:
         """Return selected clients that received the current broadcast sent by :meth:`server_broadcast`."""
         return [client for client in selected_clients if network.server() in client.messages]
-
-    def _clear_buffered_client_server_messages(
-        self,
-        network: FedNetwork,
-        selected_clients: Sequence["Agent"],
-    ) -> None:
-        """
-        Remove stale server-to-client messages for the current selected clients.
-
-        This ensures selected clients participate only if they receive the current server broadcast. The clean-up is
-        needed when :class:`~decent_bench.networks.Network` uses ``buffer_messages=True``, since old server broadcasts
-        would otherwise remain stored at clients and could be mistaken for fresh round broadcasts.
-        """
-        for client in selected_clients:
-            client._received_messages.pop(network.server(), None)  # noqa: SLF001
-
-    def _clear_buffered_server_messages(self, network: FedNetwork, participating_clients: Sequence["Agent"]) -> None:
-        """
-        Remove stale client-to-server messages for the current participants.
-
-        This ensures the server aggregates only updates actually received from those clients in the current round.
-        The clean-up is needed when :class:`~decent_bench.networks.Network` uses ``buffer_messages=True``, since old
-        client uploads would otherwise remain stored at the server and could be mistaken for fresh round updates.
-        Call this immediately before the current client upload phase.
-
-        """
-        for client in participating_clients:
-            network.server()._received_messages.pop(client, None)  # noqa: SLF001
 
     @staticmethod
     def _get_server_broadcast(client: "Agent", server: "Agent") -> "Array":
@@ -138,10 +109,6 @@ class FedAlgorithm(Algorithm[FedNetwork]):
         Aggregate client model uploads at the server using uniform averaging.
 
         This default federated aggregation assumes clients upload final local model states.
-
-        When used with :class:`~decent_bench.networks.Network` ``buffer_messages=True``, this method assumes the
-        caller has already removed stale buffered client-to-server messages for the participating clients, so only
-        current-round updates are aggregated.
         """
         received_clients = [client for client in participating_clients if client in network.server().messages]
         if not received_clients:
--- a/decent_bench/algorithms/federated/_fed_algorithm.py
+++ b/decent_bench/algorithms/federated/_fed_algorithm.py
@@ -16,7 +16,7 @@ class FedAlgorithm(Algorithm[FedNetwork]):
     """Federated algorithm - clients collaborate via a central server."""
 
     selection_scheme: ClientSelectionScheme | None = None
-    num_local_steps: LocalSteps
+    num_local_steps: LocalSteps = 1
 
     def cleanup_agents(self, network: FedNetwork) -> Iterable["Agent"]:
         return [network.server(), *network.clients()]
@@ -36,6 +36,8 @@ class FedAlgorithm(Algorithm[FedNetwork]):
             return
         if isinstance(self.num_local_steps, dict):
             for step in self.num_local_steps.values():
+                if not isinstance(step, int):
+                    raise TypeError("`num_local_steps` mapping values must be integers")
                 if step <= 0:
                     raise ValueError("`num_local_steps` must have positive values")
             return
--- a/decent_bench/algorithms/federated/_fed_algorithm.py
+++ b/decent_bench/algorithms/federated/_fed_algorithm.py
@@ -21,57 +21,6 @@ class FedAlgorithm(Algorithm[FedNetwork]):
     def cleanup_agents(self, network: FedNetwork) -> Iterable["Agent"]:
         return [network.server(), *network.clients()]
 
-    def server_broadcast(self, network: FedNetwork, selected_clients: Sequence["Agent"]) -> None:
-        """Send the current server model to the selected clients."""
-        network.send(sender=network.server(), receiver=selected_clients, msg=network.server().x)
-
-    def _clients_with_server_broadcast(self, network: FedNetwork, selected_clients: Sequence["Agent"]) -> list["Agent"]:
-        """Return the selected clients that actually received the current server broadcast."""
-        return [client for client in selected_clients if network.server() in client.messages]
-
-    def _clear_buffered_server_messages(self, network: FedNetwork, participating_clients: Sequence["Agent"]) -> None:
-        """
-        Remove stale client-to-server messages for the current participants.
-
-        This ensures the server aggregates only updates actually received from those clients in the current round.
-        The clean-up is needed when :class:`~decent_bench.networks.Network` uses ``buffer_messages=True``, since old
-        client uploads would otherwise remain stored at the server and could be mistaken for fresh round updates.
-        """
-        for client in participating_clients:
-            network.server()._received_messages.pop(client, None)  # noqa: SLF001
-
-    @staticmethod
-    def _get_server_broadcast(client: "Agent", server: "Agent") -> "Array":
-        """
-        Return the current server broadcast received by the client.
-
-        Raises:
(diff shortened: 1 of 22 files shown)


decent_bench/algorithms/federated/_fed_algorithm.py
```diff
@@ -15,29 +16,71 @@ class FedAlgorithm(Algorithm[FedNetwork]):
...
+        if isinstance(self.num_local_steps, int):
+            if self.num_local_steps <= 0:
+                raise ValueError("`num_local_steps` must be positive")
+            return
+        if isinstance(self.num_local_steps, dict):
+            for step in self.num_local_steps.values():
+                if step <= 0:
+                    raise ValueError("`num_local_steps` must have positive values")
+            return
+        raise TypeError("`num_local_steps` must be an int or a mapping from Agent to integer values")
```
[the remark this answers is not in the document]
nicola-bastianello: could also silently convert to int, but probably raising is better

decent_bench/algorithms/federated/_fed_nova.py
```diff
@@ -193,14 +169,13 @@ def step(self, network: FedNetwork, iteration: int) -> None:
...
-        self._clear_buffered_server_messages(network, participating_clients)
         self._run_local_updates(network, participating_clients)
         self._collect_received_normalizers(network, participating_clients)
         if not network.server().aux_vars["received_a_i"]:
             for client in participating_clients:
                 client.aux_vars.pop("_fednova_cumulative_gradient", None)
             return
-        self._clear_buffered_server_messages(network, participating_clients)
+        network._clear_received_messages(receivers=[network.server()], senders=participating_clients)  # noqa: SLF001
         self._communicate_cumulative_gradients(network, participating_clients)
```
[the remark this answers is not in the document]
nicola-bastianello: I would not make it part of the api, since anyway this behavior will change when addressing #302

decent_bench/algorithms/federated/_fed_algorithm.py
```diff
@@ -15,29 +16,71 @@ class FedAlgorithm(Algorithm[FedNetwork]):
     """Federated algorithm - clients collaborate via a central server."""
 
     selection_scheme: ClientSelectionScheme | None = None
+    num_local_steps: LocalSteps
```
nicola-bastianello: maybe this should default to 1 (if a fed algorithm doesn't allow for the option of multiple steps; I don't think such an algorithm exists but better safe than sorry)

docs/source/references.bib
```diff
@@ -102,6 +102,28 @@ @misc{Alg_FedLT
       note={arXiv:2511.20220 [cs]},
 }
 
+@misc{Alg_FedPD,
+      title={FedPD: A Federated Learning Framework with Optimal Rates and Adaptivity to Non-IID Data},
```
nicola-bastianello: better to use the citation to the published version https://ieeexplore.ieee.org/abstract/document/9556559

decent_bench/algorithms/federated/_fed_dyn.py
```diff
@@ -0,0 +1,150 @@
...
+    Federated Dynamic Regularization (FedDyn) with local gradient steps :footcite:p:`Alg_FedDyn`.
+
+    FedDyn keeps one dynamic state :math:`g_i` per client and one server auxiliary vector :math:`h`. In each
+    communication round, selected clients approximately minimize the dynamic-regularized local objective
+
+    .. math::
+        f_i(\theta) - \langle g_i, \theta \rangle
+        + \frac{\alpha}{2}\|\theta - \theta^t\|^2
+
+    by running ``num_local_epochs`` gradient steps from the received server model, with local gradient
```
nicola-bastianello: where is this specified in the paper? I wasn't able to find it, the pseudocode just says that the clients solve the argmin problem. this is of course impossible, so an approximate minimization like you do is the way to go. but maybe a note about this could be added

decent_bench/algorithms/federated/_fed_lt.py
```diff
@@ -50,8 +50,9 @@ class FedLT(FedAlgorithm):
     mini-batch sampling, so this behaves as local SGD. Generic :class:`~decent_bench.costs.Cost` objects use their
     normal full-gradient behavior, so this behaves as local GD.
 
-    If ``use_acceleration`` is true, Fed-LT uses the accelerated local update with each client cost's ``m_smooth`` and
-    ``m_cvx`` metadata. It initializes :math:`u_i^0=w^0_{i,k}`. With :math:`L_i=\texttt{m_smooth}` and
+    The ``local_solver`` argument selects the method used for these local steps. The default ``"gd"`` uses the
+    gradient update above. The ``"accelerated"`` option uses each client cost's ``m_smooth`` and ``m_cvx`` metadata.
```
nicola-bastianello: I would rename it to "nesterov" instead of "accelerated", since the latter is vague

decent_bench/algorithms/federated/_fed_lt.py
```diff
@@ -50,8 +50,9 @@ class FedLT(FedAlgorithm):
...
+    It initializes :math:`u_i^0=w^0_{i,k}`. With :math:`L_i=\texttt{m_smooth}` and
```
nicola-bastianello: I would organize the docstring with some titles to divide nesterov and adam, otherwise it's a single block of text

also: we should have two args, `local_solver` and `solver_args`, with the latter a dict. this way we don't have explicit solver arguments in the class signature, which I don't like

as it stands, step_size is only used for gd and adam. what I would do: allow to define a custom step-size for nesterov as well (otherwise step_size would need to be passed in the solver_args dict); but to keep it in line with all other methods that use step_size, I'm ok with having it explicit. I would also allow passing the momentum coefficient for nesterov, as a value in the solver_args dict

nicola-bastianello: as a summary:

1. have nesterov use step_size instead of 1/L+1/rho
2. pass all other solver-specific arguments in solver_args; these are momentum coefficient of nesterov, and beta1, beta2, epsilon for adam

nicola-bastianello: for nesterov, the hyperparameters currently used might not always be the best (they are the best in a centralized setting only); so having the user choose them on purpose is likely the best option. in the docstring we can mention the current values as options, but without imposing them in the code

for adam also we can mention possible values (the current defaults)

decent_bench/algorithms/federated/_fed_lt.py
```diff
@@ -132,10 +165,11 @@ def _initialize_accelerated_parameters(self, network: FedNetwork) -> None:
             m_cvx = client.cost.m_cvx
             if not math.isfinite(m_smooth) or not math.isfinite(m_cvx) or m_smooth < 0 or m_cvx < 0:
```
nicola-bastianello: these checks on m_smooth and m_cvx are to be removed since nesterov hyperparameters are chosen by the user

decent_bench/algorithms/federated/_fed_pd.py
```diff
@@ -0,0 +1,181 @@
...
+        FedPD aggregates centre candidates through :meth:`aggregate_and_synchronize`.
+
+        Raises:
+            NotImplementedError: always, because FedPD aggregation also synchronizes the updated centre.
+
+        """
+        del network, participating_clients
+        raise NotImplementedError("FedPD uses `aggregate_and_synchronize` instead of `aggregate`")
+
+    def aggregate_and_synchronize(
```
nicola-bastianello: why do we need this to be a new method instead of overwriting aggregate like in all other algorithms?

nicola-bastianello (commented):
looks good, thank you! just some relatively minor comments

decent_bench/algorithms/federated/_fed_pd.py
adrianardv: I renamed it because FedPD communicaton round is not only server side aggregation. When communication happens, the server aggregates and immediately synchronizes the resulting model back to the clients.
I wanted the method name to make that "side effect" explicit.

But if you want I can rename it back to aggregate() and document in the method docstring that FedPD's aggregate step also broadcast/synchronizes.

decent_bench/algorithms/federated/_fed_lt.py
adrianardv: okey, I made the changes. Everything should now be implemented following these comments

decent_bench/algorithms/federated/_fed_dyn.py
adrianardv: Yes. They state the client update as an exact argmin.
Then, in the experiments section of the paper, they say that for fairness, they adapt the usual SGD procedure for FedDyn local update.
I also checked their github repository, and there they implement SGD as the local solver.

I'll change a bit the docstring to state this

decent_bench/algorithms/federated/_fed_pd.py
nicola-bastianello: yes, I think it would be better to name it aggregate to keep it aligned with the other algorithms, with good documentation

decent_bench/algorithms/federated/_fed_lt.py
nicola-bastianello, line 43: I think some more organization is needed: I would put the "local solvers" title here, saying that the clients approximate  the minimization of f_i+regularization term; then three lower level titles one for each of the solvers. in each section, we should also mention clrearly what the expected solver_args are

decent_bench/algorithms/federated/_fed_lt.py
```diff
@@ -135,19 +139,38 @@ def __post_init__(self) -> None:
...
-        if self.adam_epsilon <= 0:
-            raise ValueError("`adam_epsilon` must be positive")
+        if self.local_solver not in {"gd", "nesterov", "adam"}:
+            raise ValueError("`local_solver` must be one of 'gd', 'nesterov', or 'adam'")
+        self._validate_solver_args()
+
+    def _validate_solver_args(self) -> None:
+        if self.local_solver == "adam":
+            allowed_args = {"beta1", "beta2", "epsilon"}
+            self.solver_args = {"beta1": 0.9, "beta2": 0.999, "epsilon": 1e-8, **self.solver_args}
```
nicola-bastianello: I don't like this way of creating the dict honestly. I would create `self.solver_args = {}` and then fill it in from either user passed args or from the default, raising if user-passed args are outside the range. this would make the code more readable and maintanable

decent_bench/algorithms/federated/_fed_lt.py
```diff
@@ -230,29 +230,31 @@ def _compute_local_update(self, client: "Agent", server: "Agent") -> tuple["Arra
...
         for _ in range(self.num_local_epochs):
             grad = client.cost.gradient(local_x) + ((local_x - v) / self.rho)
-            u_next = local_x - ((1 / smoothness) * grad)
+            u_next = local_x - (self.step_size * grad)
             local_x = u_next + (momentum * (u_next - u_previous))
             u_previous = u_next
         return local_x
 
     def _compute_adam_local_update(self, client: "Agent", local_x: "Array", v: "Array") -> "Array":
+        beta1 = self.solver_args["beta1"]
```
nicola-bastianello, line 261: while making the code slimmer, it might be worth it to directly access the values of the dict, without creating new references to them. I don't know if it brings much of a performance benefit, but maybe

nicola-bastianello (commented):
looks good, thank you! just a few more minor comments

decent_bench/algorithms/federated/_fed_lt.py
adrianardv: I changed it to directly accessing the dict in single-use cases (in _validate_solver_args).
For the Adam update method, though, I think keeping the local aliases is better. Since beta1 is used repeatedly in the loop, it improves readability of the formula. Also, If i'm not wrong, local variable lookup is generally cheaper than repeated dict lookup.
But if you prefer, I can also make it consistent and use direct dict access everywhere

nicola-bastianello: ok, let's keep it this way then!

nicola-bastianello:
great, thank you! merging now

PR #310, opened 2026-05-11
enh(decent-bench): fix several issues and make small enhancements

### LINKED ISSUE #268: Agent ID's
Move id assignments from the Agent creation to network initialisation to make it easier to create benchmark problems and to assure that id's are unique.

### LINKED ISSUE #285: Utilities for metric results
From experiments with the framework I’ve noticed that I tend to use the metric result object a lot for automatic tuning and other analysis. This made me wonder if we should create some utilities to help users work with this object. My first idea is a `to_dataframe` method which returns a pandas dataframe of the table results. This can be useful to quickly compare the algorithms and its a bit easier to use compared to the raw mapping structure, you can easily sort and plot subsets using the dataframe. The returned dataframe would have columns: algorithm.name, metric1, metric1_std, metric2, metric2_std, … 

Lets discuss if this is useful and if you have any other ideas.

### LINKED ISSUE #288: Algorithm initialization memory issue
When initializing algorithms using the implemented helper functions, a `dict[Agent, Array]` object is created. This causes the pickled object of the algorithm to include all agents and their cost functions. If a cost function contains large amounts of data then this causes the pickled object to become very large and takes a long time to pickle (especially unnecessary when saving initial algorithms in a checkpoint). Consider changing it to `dict[int, Array]` where the `int` is the agent id. I believe that the dict still can be accessed using the agent objects if the appropriate dunder methods are implemented on the agent. To test this issue, use the following snippet:
```python
import pickle
...

x0 = uniform_initialization(..) # or any other init
alg = DGD(x0=x0)
data = pickle.dumps(DGD)
print(f"Size of pickle with dict: {len(data) / (1024*1024)} mb")

x0 = uniform_initialization(..) # or any other init
x0 = x0.pop(0) # Get first Array, not a dict
alg = DGD(x0=x0)
data = pickle.dumps(DGD)
print(f"Size of pickle with single Array: {len(data) / (1024*1024)} mb")
```

### LINKED ISSUE #293: Messages sent
Currently messages sent are incremented by one each time a message is sent. It might be more informative to increment it by the size of the message to better reflect differences in message size. Some methods might send larger or smaller messages. It is debatable if this should also be updated for the number of dropped and received messages.

### LINKED ISSUE #296: More efficient noise generation
Currently, noise is generated inside `send_one`, which is not very efficient since it requires the creation of several arrays of noise. A more efficient way to do this would be to create a big array with all the noise in `send`, and then a portion of this is used for each of the communications. In particular, this could look like this:

- we modify noise schemes to take a tuple representing the shape of the noise array to be created (instead of passing the message like we do now)
- the noise array then is created in `send` as `(d_i,) + cost.shape` where d_i is the number of messages to be transmitted
- we add the noise to the message and only then pass it to `send_one`

One issue I see: if there is no noise, we can pass an array of zeros, but that requires creating this array and then summing it to the message, which is a lot of wasted computation. So I think the noise scheme should return Array | None, and if `send` gets None then we just pass through the original message

### LINKED ISSUE #299: Custom __add__ for empirical risk costs yields wrong x_optimal
I was running some tests and realized that the regret was coming out negative, when it should always be nonnegative. I traced this back to the use of `sum_cost = reduce(add, classification_costs)` in `create_classification_problem`. indeed, this calls the custom `__add__` in `LogisticRegressionCost` and minimizes the result, but this is different from minimizing `Sum(classification_costs)`, which is actually what we want

the problem is: given `f` and `g` logistic costs with `n_f` and `n_g` data, if we do `sum_cost = reduce(add, [f, g])` we get a new logistic cost with `n_f+n_g` data, which means that its function and gradient are divided by `n_f+n_g`. instead if we do `Sum([f, g])`, we get a cost of the type (1/n_f) sum_{h=1}^{n_f} l(x, d_h^f) + (1/n_g) sum_{h=1}^{n_g} l(x, d_h^g) which correctly preserves the scaling of the component costs f and g. in other words, if we use the custom `__add__` we are not minimizing the correct weighted sum of f+g

now, I see three options to solve this:

1. we switch to `Sum(classification_costs)` in `create_classification_problem`; this should ensure the correct cost is being minimized. however, it still leaves the door open for the same issue to happen again if users go through `__add__`
2. we remove the custom `__add__` from LogisticRegressionCost and LinearRegressionCost, and encourage users to do `LogisticRegressionCost(dataset=dataset_f + dataset_g)` if they want the same result as the current custom `__add__`. the problem with this is that we lose a lot of efficiency because computing gradient of SumCost([f, g]) is less efficient than f+g using custom `__add__`
3. we change the definition of empirical risk costs, removing the division by the number of datapoints. this means that the custom `__add__` will coincide with the expected behavior of f+g, while preserving efficiency. the problem here is that values of function/gradient will quickly become quite large

none of these options is perfect, but probably 2. is the one more in line with what people expect. and even if we lose some efficiency, we don't have any surprising behavior

(another related issue is that `create_classification_problem` doesn't use `indices="all"` when computing x_optimal for empirical costs, but that's an easy fix)

nicola-bastianello:
the PR addresses a few issues and adds a few small enhancements.

1. agent id:
    - agents now have a unique id assigned at init using uuid; this is used as the hash of the agent and is immutable
    - the constructor no longer requires an id
    - the agents also have an index (which implements `__index__`) that is-1 at init and assigned by a network; the network init raises if an agent has index != -1 (it's already assigned). the index is mutable
    - id was also added to AgentMetricsView (in preparation for addressing #313 )
2. networks:
    - P2PNetwork can now only be constructed from a list of agents and nx.Graph (nx.Graph[Agent] not allowed anymore)
    - _send_one was removed, centralizing all functionality in send to make noise generation more efficient
    - message counters are now updated according to their shape; in particular, counters are updated as the fraction of mssage size to cost shape. added `compressed_msg_size` to compression schemes to account for the reduction in size that would occur in real life
    - plot_networks now takes directly Network object instead of nx.Graph
3. schemes: changed Quantization to uniform quantization, which allows using much more efficient numpy function `rint`
4. metrics computation/display pipeline:
    - table_description and plot_description unified in `description` property
    - removed the option to pass a list of lists for plot_metrics, simplifying internal logic significantly; option for individual_plots is still available, and still max 3 subplots per figure
    - simplified some minor logic here and there
    - added `to_dataframe` to `MetricResult`
5. statistics:
    - statistics is now an optional arg, defaulting to None; in that case, it applies `default_statistic`, which is average if the get_trial_data returns a list, or the single value otherwise
    - table header specifies "statistic across agents"
    - removed utils.single, since the default None takes care of that through the new `default_statistic`; passing an empty list also defaults to `default_statistic`
    - table lables only have "(statistic)" if specified by the user
6. metrics:
    - XError now computes w.r.t. to the mean; this is more efficent (since x_mean is cached) and provides basically the same info
    - Regret and GradientNorm use batch=all when applied to empirical costs that admit a x_optimal (logistic and linear regression)
    - GradientNorm: removed squaring the norm
    - Regret: divided by the num. of agents
7. create problems:
    - added option to avoid computing x_optimal
    - custom branch for Sum(LinearRegressionCost) and custom x_optimal computation in create_classification_problem
8. storing initial algorithms: now all Agent-keyed dicts are stored replacing the Agent instance with its hash, and restoring to Agent at unpickle
9. logistic and linear regression:
    - removed custom add (which had the potential to give incorrect results e.g. when computing x_optimal for costs with different n_samples
    - make _get_batch_data more efficient
 
closes #312 closes #299 closes #296 closes #293 closes #288 closes #285 closes #268

### DIFF
--- a/decent_bench/benchmark/_metric_result.py
+++ b/decent_bench/benchmark/_metric_result.py
@@ -114,13 +114,15 @@ class MetricResult:
             for algorithm, table_metric_results in self.table_results.items():
                 for metric, statistics in table_metric_results.items():
                     for statistic_name, (mean, margin_of_error) in statistics.items():
-                        table_rows.append({
-                            "algorithm": algorithm.name,
-                            "metric": metric.description,
-                            "statistic": statistic_name,
-                            "mean": mean,
-                            "margin_of_error": margin_of_error,
-                        })
+                        table_rows.append(
+                            {
+                                "algorithm": algorithm.name,
+                                "metric": metric.description,
+                                "statistic": statistic_name,
+                                "mean": mean,
+                                "margin_of_error": margin_of_error,
+                            }
+                        )
 
             table_frame = pd.DataFrame.from_records(
                 table_rows,
@@ -139,14 +141,16 @@ class MetricResult:
                         y_max_values,
                         strict=True,
                     ):
-                        plot_rows.append({
-                            "algorithm": algorithm.name,
-                            "metric": metric.description,
-                            "x": x_value,
-                            "y_mean": y_mean,
-                            "y_min": y_min,
-                            "y_max": y_max,
-                        })
+                        plot_rows.append(
+                            {
+                                "algorithm": algorithm.name,
+                                "metric": metric.description,
+                                "x": x_value,
+                                "y_mean": y_mean,
+                                "y_min": y_min,
+                                "y_max": y_max,
+                            }
+                        )
 
             plot_frame = pd.DataFrame.from_records(
                 plot_rows,
--- a/decent_bench/benchmark/_metric_result.py
+++ b/decent_bench/benchmark/_metric_result.py
@@ -5,7 +5,6 @@ from decent_bench.agents import AgentMetricsView
 from decent_bench.algorithms import Algorithm
 from decent_bench.metrics import Metric
 from decent_bench.networks import Network
-from decent_bench.utils._metric_helpers import _flatten_plot_metrics
 
 
 @dataclass
@@ -33,7 +32,7 @@ class MetricResult:
 
     agent_metrics: Mapping[Algorithm[Network], Sequence[Sequence[AgentMetricsView]]] | None
     table_metrics: list[Metric] | None
-    plot_metrics: list[Metric] | list[list[Metric]] | None
+    plot_metrics: list[Metric] | None
     table_results: Mapping[Algorithm[Network], Mapping[Metric, Mapping[str, tuple[float, float]]]] | None
     plot_results: (
         Mapping[
@@ -55,10 +54,10 @@ class MetricResult:
 
     @property
     def available_table_metrics(self) -> list[str]:
-        """Return ``table_description`` of available metrics, which can be used for filtering in :func:`~decent_bench.benchmark.display_metrics`."""  # noqa: E501
-        return sorted({metric.table_description for metric in (self.table_metrics or [])})
+        """Return ``description`` of available table metrics, which can be used for filtering in :func:`~decent_bench.benchmark.display_metrics`."""  # noqa: E501
+        return sorted({metric.description for metric in (self.table_metrics or [])})
 
     @property
     def available_plot_metrics(self) -> list[str]:
-        """Return ``plot_descriptions`` of available metrics, which can be used for filtering in :func:`~decent_bench.benchmark.display_metrics`."""  # noqa: E501
-        return sorted({metric.plot_description for metric in (_flatten_plot_metrics(self.plot_metrics or []))})
+        """Return ``description`` of available plot metrics, which can be used for filtering in :func:`~decent_bench.benchmark.display_metrics`."""  # noqa: E501
+        return sorted({metric.description for metric in (self.plot_metrics or [])})
--- a/decent_bench/benchmark/_metric_result.py
+++ b/decent_bench/benchmark/_metric_result.py
@@ -1,23 +1,12 @@
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
-from importlib import import_module
-from types import ModuleType
-from typing import TYPE_CHECKING
+
+import pandas as pd
 
 from decent_bench.agents import AgentMetricsView
 from decent_bench.algorithms import Algorithm
 from decent_bench.metrics import Metric
 from decent_bench.networks import Network
-
-if TYPE_CHECKING:
-    import pandas
-
-
-def _import_pandas() -> ModuleType:
-    try:
-        return import_module("pandas")
-    except ImportError as exc:  # pragma: no cover - exercised when pandas is unavailable
-        raise ImportError("pandas is required to convert MetricResult to dataframes") from exc
 
 
 @dataclass
@@ -82,7 +71,7 @@ class MetricResult:
         """Return ``description`` of available plot metrics, which can be used for filtering in :func:`~decent_bench.benchmark.display_metrics`."""  # noqa: E501
         return sorted({metric.description for metric in (self.plot_metrics or [])})
 
-    def to_dataframe(self) -> tuple["pandas.DataFrame | None", "pandas.DataFrame | None"]:
+    def to_dataframe(self) -> tuple["pd.DataFrame | None", "pd.DataFrame | None"]:
         """
         Convert the stored metric results into pandas dataframes.
 
@@ -106,8 +95,6 @@ class MetricResult:
                 y_mean_values = loss_rows["y_mean"]
 
         """
-        pd = _import_pandas()
-
         table_frame = None
         if self.table_results is not None:
             table_rows: list[dict[str, object]] = []
--- a/decent_bench/benchmark/_metric_result.py
+++ b/decent_bench/benchmark/_metric_result.py
@@ -10,7 +10,7 @@ from decent_bench.metrics import Metric
 from decent_bench.networks import Network
 
 if TYPE_CHECKING:
-    from pandas import DataFrame
+    import pandas
 
 
 def _import_pandas() -> ModuleType:
@@ -75,7 +75,7 @@ class MetricResult:
         """Return ``description`` of available plot metrics, which can be used for filtering in :func:`~decent_bench.benchmark.display_metrics`."""  # noqa: E501
         return sorted({metric.description for metric in (self.plot_metrics or [])})
 
-    def to_dataframe(self) -> tuple["DataFrame | None", "DataFrame | None"]:
+    def to_dataframe(self) -> tuple["pandas.DataFrame | None", "pandas.DataFrame | None"]:
         """
         Convert the stored metric results into pandas dataframes.
 
@@ -104,18 +104,16 @@ class MetricResult:
         table_frame = None
         if self.table_results is not None:
             table_rows: list[dict[str, object]] = []
-            for algorithm, metric_results in self.table_results.items():
-                for metric, statistics in metric_results.items():
+            for algorithm, table_metric_results in self.table_results.items():
+                for metric, statistics in table_metric_results.items():
                     for statistic_name, (mean, margin_of_error) in statistics.items():
-                        table_rows.append(
-                            {
-                                "algorithm": algorithm.name,
-                                "metric": metric.description,
-                                "statistic": statistic_name,
-                                "mean": mean,
-                                "margin_of_error": margin_of_error,
-                            }
-                        )
+                        table_rows.append({
+                            "algorithm": algorithm.name,
+                            "metric": metric.description,
+                            "statistic": statistic_name,
+                            "mean": mean,
+                            "margin_of_error": margin_of_error,
+                        })
 
             table_frame = pd.DataFrame.from_records(
                 table_rows,
@@ -125,8 +123,8 @@ class MetricResult:
         plot_frame = None
         if self.plot_results is not None:
             plot_rows: list[dict[str, object]] = []
-            for algorithm, metric_results in self.plot_results.items():
-                for metric, (x_values, y_mean_values, y_min_values, y_max_values) in metric_results.items():
+            for algorithm, plot_metric_results in self.plot_results.items():
+                for metric, (x_values, y_mean_values, y_min_values, y_max_values) in plot_metric_results.items():
                     for x_value, y_mean, y_min, y_max in zip(
                         x_values,
                         y_mean_values,
@@ -134,16 +132,14 @@ class MetricResult:
                         y_max_values,
                         strict=True,
                     ):
-                        plot_rows.append(
-                            {
-                                "algorithm": algorithm.name,
-                                "metric": metric.description,
-                                "x": x_value,
-                                "y_mean": y_mean,
-                                "y_min": y_min,
-                                "y_max": y_max,
-                            }
-                        )
+                        plot_rows.append({
+                            "algorithm": algorithm.name,
+                            "metric": metric.description,
+                            "x": x_value,
+                            "y_mean": y_mean,
+                            "y_min": y_min,
+                            "y_max": y_max,
+                        })
 
             plot_frame = pd.DataFrame.from_records(
                 plot_rows,
--- a/decent_bench/benchmark/_metric_result.py
+++ b/decent_bench/benchmark/_metric_result.py
@@ -1,10 +1,23 @@
 from collections.abc import Mapping, Sequence
 from dataclasses import dataclass
+from importlib import import_module
+from types import ModuleType
+from typing import TYPE_CHECKING
 
 from decent_bench.agents import AgentMetricsView
 from decent_bench.algorithms import Algorithm
 from decent_bench.metrics import Metric
 from decent_bench.networks import Network
+
+if TYPE_CHECKING:
+    from pandas import DataFrame
+
+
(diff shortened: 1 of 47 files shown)


docs/source/user.rst
adrianardv: This example is using the old `Agent(agent_id, cost)`

adrianardv: Nested plot_metrics have been removed. This should be updated

adrianardv: Same here. n_significant_digits is now outdated

adrianardv: Also outdated, right?

decent_bench/networks.py
```diff
@@ -289,14 +293,29 @@ def send(
...
-        currently_active_connected_agents = self._allowed_receivers(sender)
+        # raise for unconnected/inactive
+        unavailable_missing_receivers = set(receiver) - self._allowed_receivers(sender)
+        if unavailable_missing_receivers:
+            raise ValueError(
+                f"Receivers {unavailable_missing_receivers} are not active or not connected to {sender} "
+                f"in iteration {self._iteration}"
+            )
+
+        counter_increment = int(np.prod(iop.shape(msg)) / sender.cost.size)  # replace with msg.size once available
```
adrianardv: Should `counter_increment`be cast to int here?

I understand the idea of normalizing by message size/model size, but with the int() cast, messages smaller than `sender.cost.size` would contribute 0 to the counters.

Would it make sense to keep this value fractional, so scalar o otherwise smaller than model messages are still accounted for?

adrianardv (commented):
Looks good. Just a few minor comments

decent_bench/networks.py
nicola-bastianello: I was thinking about it, and I decided to go for int because 1) it was typed as such in several places, 2) I'm not sure any messages are transmitted that don't have size a multiple of the cost.shape, or at least the vast majority are

I'm not against making it fractional like gradient evaluation, I just was thinking it might not be necessary. have you encountered many algorithms that use smaller than model-size messages?

Simpag: I will take a look at this PR during the weekend, but a quick observation here. Algorithms might do some compression that reduces packet size at send and uncompress at receive. These could be much smaller than cost.shape. To further support this we might add a uncompress method to compression schemes

nicola-bastianello: good idea @Simpag , I'll open an issue about compress/uncompress. and I'll make the counter fractional

adrianardv: > I was thinking about it, and I decided to go for int because 1) it was typed as such in several places, 2) I'm not sure any messages are transmitted that don't have size a multiple of the cost.shape, or at least the vast majority are
> 
> I'm not against making it fractional like gradient evaluation, I just was thinking it might not be necessary. have you encountered many algorithms that use smaller than model-size messages?

In the currently implemented algorithms, FedNova sends the scalar normalizers. 

I'm not sure if other algorithms which currently send stacked "payloads" containing multiple variables, might send "separate" smaller than sender.cost.size messages when #302 is tackled.

Simpag: Compression schemes with lossless compression would be a waste of compute but could be interesting to implement some lossy compression schemes. If we want to save some compute if they do numpy conversions we could make the compress method return the compressed size which will be used for ”transmission cost” and not have an uncompressed method

nicola-bastianello: I agree, actually performing compress/uncompress is wasting compute. I also landed on the option of just updating the counter (e.g. with a new attribute `CompressionScheme.compression_rate`, see #312. I will address it in this PR

decent_bench/benchmark/_metric_result.py
```diff
@@ -1,11 +1,23 @@
...
+
+if TYPE_CHECKING:
+    import pandas
+
+
+def _import_pandas() -> ModuleType:
+    try:
+        return import_module("pandas")
+    except ImportError as exc:  # pragma: no cover - exercised when pandas is unavailable
+        raise ImportError("pandas is required to convert MetricResult to dataframes") from exc
```
Simpag: Can just import pandas normally and we add it as a dep since we will eventually use it more

decent_bench/benchmark/_metric_result.py
```diff
@@ -62,10 +74,87 @@ def available_algorithms(self) -> list[str]:
...
+                        strict=True,
+                    ):
+                        plot_rows.append(
+                            {
+                                "algorithm": algorithm.name,
+                                "metric": metric.description,
+                                "x": x_value,
+                                "y_mean": y_mean,
+                                "y_min": y_min,
+                                "y_max": y_max,
```
Simpag, line 138: Maybe unpack the values, add a column called iteration and have one row of values for each iteration. Will be some duplicated data but thats fine imo. Then you can make plots using plot_frame.plot() and such. Might be possible using the lists, not sure

decent_bench/benchmark/_utils.py
```diff
@@ -99,22 +104,21 @@ def model_gen() -> torch.nn.Module:
...
-        x_optimal = ca.solve(
-            sum_cost,
-            max_iter=SOLVE_MAX_ITER,
-            stop_tol=SOLVE_STOP_TOL,
-            max_tol=SOLVE_MAX_TOL,
-            show_progress=show_progress,
-        )
+        if compute_x_optimal:
+            # agents have the same n_samples, so minimizing a single logistic cost with all data is equivalent
+            sum_cost = LogisticRegressionCost(dataset=dataset.get_datapoints(), batch_size="all")
```
Simpag, line 114: This works fine here, but we need a proper solution for when its not possible to do full gradient evaluations. Maybe the ca solver could take a list of costs instead so its able to properly reduce the total gradients

decent_bench/metrics/_plots.py
```diff
@@ -362,44 +358,24 @@ def _save_and_show_figures(
...
-    flat_metrics_list: list[Metric] = metrics  # type: ignore[assignment]
     if individual_plots:
-        # Each metric in its own figure
-        return [[metric] for metric in flat_metrics_list]
+        return [[metric] for metric in metrics]
 
-    # Group into chunks of up to 3 metrics per figure
-    groups: list[list[Metric]] = [flat_metrics_list[i : i + 3] for i in range(0, len(flat_metrics_list), 3)]
-    return groups
+    return [metrics[i : i + 3] for i in range(0, len(metrics), 3)]
```
Simpag: Dont hardcode the 3 here, make it a const at the top

decent_bench/metrics/_tables.py
```diff
@@ -155,6 +155,8 @@ def compute_tables(
 
             for statistic in metric.statistics:
                 statistic_name = STATISTICS_ABBR.get(statistic.__name__) or statistic.__name__
+                if statistic_name == "default_statistic":
+                    statistic_name = ""
```
Simpag, line 159: Could add it to STATISTICS_ABBR instead, maybe renaming the const to something along the lines of statistic mapping

decent_bench/metrics/metric_utils.py
Simpag: We should consider using LRU cache instead so we can set a max size. If we have very long running experiments with low snapshot period it will grow quickly

decent_bench/metrics/metric_utils.py
```diff
@@ -112,9 +105,12 @@ def _regret(agents: Sequence[AgentMetricsView], problem: "BenchmarkProblem", ite
     """
     x_opt = problem.x_optimal
     mean_x = x_mean(tuple(agents), iteration)
-    optimal_cost = sum(a.cost.function(x_opt) for a in agents)  # type: ignore[arg-type, misc]
-    actual_cost = sum(a.cost.function(mean_x) for a in agents)
-    return actual_cost - optimal_cost
+    optimal_cost, actual_cost = 0.0, 0.0
+    for a in agents:
+        kwargs = {"indices": "all"} if isinstance(a.cost, EmpiricalRiskCost) else {}
```
Simpag, line 113: Since we are using indices = all in a few places. Maybe we should require that all EmpiricalRiskCosts can handle it no matter how big the underlying dataset is, that is all empirical costs batch those operations silently. This is being done in pytorch cost with a maximum batch size

decent_bench/metrics/metric_utils.py
```diff
@@ -126,26 +122,27 @@ def _gradient_norm(agents: Sequence[AgentMetricsView], iteration: int = -1) -> f
...
+    where :math:`\mathbf{\bar{x}}_k` is the mean of the agents' local x at iteration k,
     and :math:`\mathbf{x}^\star` is the optimal x defined in the *problem*.
 
     """
-    agent_iteration = agent.x_history.max() if iteration == -1 else iteration
-    x_at_iteration = iop.to_numpy(agent.x_history[agent_iteration])
-    opt_x = iop.to_numpy(problem.x_optimal)  # type: ignore[arg-type]
-    return float(la.norm(x_at_iteration - opt_x))
+    mean_x = x_mean(tuple(agents), iteration)
+    return float(la.norm(iop.to_numpy(mean_x - problem.x_optimal)))  # type: ignore[operator]
```
Simpag: This could be performed using iop

decent_bench/networks.py
```diff
@@ -67,10 +64,13 @@ def __init__(
...
             raise ValueError("The graph needs to be connected")
-        agent_ids = [agent.id for agent in graph.nodes()]
-        if len(agent_ids) != len(set(agent_ids)):
-            raise ValueError("Agent IDs must be unique")
+        if any(not isinstance(node, Agent) for node in graph):
+            raise ValueError("The graph nodes must be `Agent` objects")
         self._validate_agent_cost_compatibility(graph)
+        for idx, agent in enumerate(graph.nodes()):  # assign agent index within the network
+            if agent.index != -1:
+                raise ValueError("Agents can only be assigned to one network at a time")
```
Simpag: Id do a check first so you dont assign id's on some agents and then fail, can just use the static method

decent_bench/benchmark/_metric_result.py
nicola-bastianello: I've waited to make changes to this, will do them in #309

decent_bench/benchmark/_utils.py
nicola-bastianello: in other case the ca solver would get a SumCost, which contains indeed the list of costs. but I'm not sure I understand which problem you refer to

decent_bench/metrics/_tables.py
nicola-bastianello: I'll leave it like this for now since it's a temporary solution until #309 is addressed

decent_bench/metrics/metric_utils.py
nicola-bastianello: good point! I don't have a good sense for what a good cache size is though. any ideas?

decent_bench/metrics/metric_utils.py
nicola-bastianello: the "all" option is used only when the functions admit an optimal solution, so if I'm not mistaken PyTorchCost would never be passed such an option. but yes, I'll make a note to look into this in the future

decent_bench/metrics/metric_utils.py
nicola-bastianello: I used 50 for now, maybe later we can do some tests to figure out a good size

decent_bench/benchmark/_utils.py
Simpag: The problem is that lets say you have a cost function with 1M datapoints, its not possible to calculate the entire "all" batch. It wont fit into memory, then you'd have to either batch it on the solver side or have "silent" batching on the cost side. I would say that cost functions should silently do this batching. The full gradient is equivalent to the mean of the gradient on all datapoints so it doesn’t change the solution

Simpag: So when a cost function receives indices = "all" it should decide how to calculate this. For the PyTorchCost it has a property called "max_batch_size" which I believe by default is set to the batch_size. If it receives indices = "all" it will silently perform mini-batches over the entire dataset and aggregate these so its transparent to the user

decent_bench/metrics/metric_utils.py
Simpag: I think 50 is way too small. We want the cache to survive between different metric calculations otherwise it doesnt help. I'd set it to around 50k. For example, I ran 10 algorithms over 20k iterations with a sample period of I believe 50. This yields 4000 datapoints per trial, with 5 trials thats 20k caches. We always flush the cache memory when its not needed anymore so its fine to have it large but its good to have a hard cap. I dont believe we store insane amounts of information in the caches, mostly small lists, so it should be fine to have it large. Its mostly to prevent memory explosions on VERY long benchmarks

decent_bench/metrics/metric_utils.py
Simpag: The all option is (and should) be used on multiple metrics so it will probably occur on all types of costs

Simpag:
After you've increased the cache size I think everything looks good

decent_bench/benchmark/_utils.py
nicola-bastianello: ok I see your point. then I agree that this should be handled in the backend. I'll make a note about it

decent_bench/metrics/metric_utils.py
nicola-bastianello: yes of course, don't know what I was thinking

PR #311, opened 2026-05-13
enh(metrics): Add optional/extra federated-specific bencmark metrics

adrianardv:
### Summary

Add federated-specific benchmark metrics so that inspection of server behavior, participation, uplink/downlink communication, and client-server drift after running federated benchmarks is possible.

The existing client-based metrics remain unchanged.

### Changes
- Add server metric snapshots for federated benchmark results.
- Expose server metric views through `MetricResult`.
- Add federated metrics for:
  - client drift from server
  - selected participation fraction
  - total communication, uplink/downlink messages, downlink messages dropped
  - server MSE
  - server accuracy
- Add/update tests for federated defaults, server metrics, and custom
  metric override behavior.

### DIFF
--- a/decent_bench/algorithms/federated/_fed_algorithm.py
+++ b/decent_bench/algorithms/federated/_fed_algorithm.py
@@ -82,7 +82,7 @@ class FedAlgorithm(Algorithm[FedNetwork]):
     def _record_selected_clients(selected_clients: Sequence["Agent"]) -> None:
         """Record clients selected by a federated round."""
         for client in selected_clients:
-            client._n_times_selected += 1  # noqa: SLF001
+            client._n_times_selected = getattr(client, "_n_times_selected", 0) + 1  # noqa: SLF001
 
     def server_broadcast(self, network: FedNetwork, selected_clients: Sequence["Agent"]) -> None:
         """Send the current server model to the selected clients."""
--- a/decent_bench/algorithms/federated/_fed_algorithm.py
+++ b/decent_bench/algorithms/federated/_fed_algorithm.py
@@ -64,7 +64,7 @@ class FedAlgorithm(Algorithm[FedNetwork]):
 
     def _selected_clients_for_round(self, network: FedNetwork, iteration: int) -> list["Agent"]:
         """
-        Select participating clients for the current round from the currently active clients.
+        Select participating clients for the current round from the currently active clients and record them.
 
         If ``self.selection_scheme`` is ``None``, all active clients are selected.
         """
--- a/decent_bench/algorithms/federated/_fed_algorithm.py
+++ b/decent_bench/algorithms/federated/_fed_algorithm.py
@@ -62,14 +62,16 @@ class FedAlgorithm(Algorithm[FedNetwork]):
             )
         return {client: self.num_local_steps[client] for client in clients}
 
-    def _selected_clients_for_round(self, network: FedNetwork, iteration: int) -> list["Agent"]:
+    def selected_clients_for_round(self, network: FedNetwork, iteration: int) -> list["Agent"]:
         """
-        Select participating clients for the current round from the currently active clients and record them.
+        Select participating clients for the current round and record participation.
+
+        Use this method once per federated round when selecting participating clients in the round. It applies
+        ``self.selection_scheme`` to the currently active clients and updates the selection counters used by
+        participation metrics. Calling it more than once in the same round will increment the selected-client counters
+        more than once.
 
         If ``self.selection_scheme`` is ``None``, all active clients are selected.
-
-        Federated algorithms should use this helper for round-level client selection so participation counters remain
-        consistent with the executed training path.
         """
         active_clients = network.active_clients()
         if not active_clients:
--- a/decent_bench/algorithms/federated/_fed_algorithm.py
+++ b/decent_bench/algorithms/federated/_fed_algorithm.py
@@ -124,14 +124,7 @@ class FedAlgorithm(Algorithm[FedNetwork]):
         received_clients = [client for client in participating_clients if client in network.server().messages]
         if not received_clients:
             return
-        self._record_updates_received_by_server(received_clients)
         updates = [network.server().messages[client] for client in received_clients]
         weights = [1.0] * len(received_clients)
         total_weight = float(len(received_clients))
         network.server().x = self._weighted_average(updates, weights, total_weight)
-
-    @staticmethod
-    def _record_updates_received_by_server(received_clients: Sequence["Agent"]) -> None:
-        """Record clients whose update contributed to a server aggregation."""
-        for client in received_clients:
-            client._n_times_update_received_by_server += 1  # noqa: SLF001
--- a/decent_bench/algorithms/federated/_fed_algorithm.py
+++ b/decent_bench/algorithms/federated/_fed_algorithm.py
@@ -67,6 +67,9 @@ class FedAlgorithm(Algorithm[FedNetwork]):
         Select participating clients for the current round from the currently active clients and record them.
 
         If ``self.selection_scheme`` is ``None``, all active clients are selected.
+
+        Federated algorithms should use this helper for round-level client selection so participation counters remain
+        consistent with the executed training path.
         """
         active_clients = network.active_clients()
         if not active_clients:
--- a/decent_bench/algorithms/federated/_fed_algorithm.py
+++ b/decent_bench/algorithms/federated/_fed_algorithm.py
@@ -62,14 +62,12 @@ class FedAlgorithm(Algorithm[FedNetwork]):
             )
         return {client: self.num_local_steps[client] for client in clients}
 
-    def selected_clients_for_round(self, network: FedNetwork, iteration: int) -> list["Agent"]:
+    def select_clients(self, network: FedNetwork, iteration: int) -> list["Agent"]:
         """
-        Select participating clients for the current round and record participation.
+        Select clients for the current federated round.
 
-        Use this method once per federated round when selecting participating clients in the round. It applies
-        ``self.selection_scheme`` to the currently active clients and updates the selection counters used by
-        participation metrics. Calling it more than once in the same round will increment the selected-client counters
-        more than once.
+        The method selects the subset of active clients that will receive the server broadcast and perform local
+        training. The clients are selected using ``self.selection_scheme``.
 
         If ``self.selection_scheme`` is ``None``, all active clients are selected.
         """
@@ -87,7 +85,7 @@ class FedAlgorithm(Algorithm[FedNetwork]):
     def _record_selected_clients(selected_clients: Sequence["Agent"]) -> None:
         """Record clients selected by a federated round."""
         for client in selected_clients:
-            client._n_times_selected = getattr(client, "_n_times_selected", 0) + 1  # noqa: SLF001
+            client._n_times_selected += 1  # noqa: SLF001
 
     def server_broadcast(self, network: FedNetwork, selected_clients: Sequence["Agent"]) -> None:
         """Send the current server model to the selected clients."""
--- a/decent_bench/algorithms/federated/_fed_algorithm.py
+++ b/decent_bench/algorithms/federated/_fed_algorithm.py
@@ -72,8 +72,17 @@ class FedAlgorithm(Algorithm[FedNetwork]):
         if not active_clients:
             return []
         if self.selection_scheme is None:
-            return list(active_clients)
-        return self.selection_scheme.select(active_clients, iteration)
+            selected_clients = list(active_clients)
+        else:
+            selected_clients = self.selection_scheme.select(active_clients, iteration)
+        self._record_selected_clients(selected_clients)
+        return selected_clients
+
+    @staticmethod
+    def _record_selected_clients(selected_clients: Sequence["Agent"]) -> None:
+        """Record clients selected by a federated round."""
+        for client in selected_clients:
+            client._n_times_selected += 1  # noqa: SLF001
 
     def server_broadcast(self, network: FedNetwork, selected_clients: Sequence["Agent"]) -> None:
         """Send the current server model to the selected clients."""
@@ -115,7 +124,14 @@ class FedAlgorithm(Algorithm[FedNetwork]):
         received_clients = [client for client in participating_clients if client in network.server().messages]
         if not received_clients:
             return
+        self._record_updates_received_by_server(received_clients)
         updates = [network.server().messages[client] for client in received_clients]
         weights = [1.0] * len(received_clients)
         total_weight = float(len(received_clients))
         network.server().x = self._weighted_average(updates, weights, total_weight)
+
+    @staticmethod
+    def _record_updates_received_by_server(received_clients: Sequence["Agent"]) -> None:
+        """Record clients whose update contributed to a server aggregation."""
+        for client in received_clients:
+            client._n_times_update_received_by_server += 1  # noqa: SLF001
--- a/decent_bench/algorithms/federated/_fed_pd.py
+++ b/decent_bench/algorithms/federated/_fed_pd.py
@@ -100,9 +100,7 @@ class FedPD(FedAlgorithm):
         self.num_local_steps = self._num_local_steps_by_client
 
     def step(self, network: FedNetwork, iteration: int) -> None:
-        if self.selection_scheme is not None:
-            raise ValueError("FedPD does not support client selection")
-        participating_clients = self.selected_clients_for_round(network, iteration)
+        participating_clients = self.select_clients(network, iteration)
         if not participating_clients:
             return
 
--- a/decent_bench/algorithms/federated/_fed_pd.py
+++ b/decent_bench/algorithms/federated/_fed_pd.py
@@ -102,7 +102,7 @@ class FedPD(FedAlgorithm):
     def step(self, network: FedNetwork, iteration: int) -> None:
         if self.selection_scheme is not None:
             raise ValueError("FedPD does not support client selection")
-        participating_clients = self._selected_clients_for_round(network, iteration)
+        participating_clients = self.selected_clients_for_round(network, iteration)
         if not participating_clients:
             return
 
--- a/decent_bench/algorithms/federated/_fed_pd.py
+++ b/decent_bench/algorithms/federated/_fed_pd.py
@@ -103,6 +103,7 @@ class FedPD(FedAlgorithm):
         participating_clients = network.active_clients()
         if not participating_clients:
             return
+        self._record_selected_clients(participating_clients)
 
         self._run_local_updates(participating_clients)
         if random.random() >= self.skip_probability:
@@ -160,6 +161,7 @@ class FedPD(FedAlgorithm):
         received_clients = [client for client in participating_clients if client in network.server().messages]
         if not received_clients:
             return
+        self._record_updates_received_by_server(received_clients)
         center_candidates = [server.messages[client] for client in received_clients]
         weights = [1.0] * len(received_clients)
         total_weight = float(len(received_clients))
--- a/decent_bench/algorithms/federated/_fed_pd.py
+++ b/decent_bench/algorithms/federated/_fed_pd.py
@@ -161,7 +161,6 @@ class FedPD(FedAlgorithm):
         received_clients = [client for client in participating_clients if client in network.server().messages]
         if not received_clients:
             return
-        self._record_updates_received_by_server(received_clients)
         center_candidates = [server.messages[client] for client in received_clients]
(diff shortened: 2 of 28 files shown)


nicola-bastianello:
thank you! I'll review this in the coming days

nicola-bastianello:
## general thoughts

it's a great idea to add these federated-specific metrics; however, the current implementation makes too extensive of a refactor, and I want to keep the metrics computation and display pipeline unchanged if possible. here are two options on how to do this:

### option 1
just like you added `_n_times_selected` to Agent, we could add `_is_server = False`. then, following the idea of `snapshot_agents` in Network, we pass to all metrics a `agents: Sequence[AgentMetricsView]` that, for FedNetwork, includes the server. then the metrics that only want clients can do `[agent for agent in agents if not agent._is_server]` (this would return all agents in P2PNetworks)

note: I think the custom `snapshot_agents` in `FedNetwork` can be made redundant by changing `Network.snapshot_agents` to return `self.graph.nodes()` instead, which includes the server

this would allow removing `get_federated_table_data` and `get_federated_plot_data`. in my opinion we should only provide one core api function to compute metrics, and that's `get_data_from_trial`. then during metric computation we don't need to have separate branches for the federated case, since all metrics are computed through the same api. `load_metrics_results` also would not need to change

### option 2
option 1 is probably the simplest for now, but this PR has made me think: it is probably not the best idea to pass a `Sequence[AgentMetricsView]` to metrics, since we might need information from the network when computing metrics (like knowing which agent is the server). a slightly bigger refactor then could be: we create a new object like `NetworkMetricsView` which holds the same topology as the originating `Network`, but replaces agents with `AgentsMetricsView` and doesn't include most network custom functions (except for neighbors, coordinator/server, agents/clients, which are useful). the change in the api would be quite minimal I think, just a arg/type change from `agents: Sequence[AgentMetricsView]` to `network: NetworkMetricsView`

this is definitely the most future-proof choice, since we don't discard network information which could be helpful as we add new Metrics (like this PR demonstrates)


## other notes

- I don't see where `_record_selected_clients` is used and for what, maybe it can be removed?
- the note "Note: For :class:`~decent_bench.networks.FedNetwork`, this metric is evaluated at the mean client model, not at the server model." on all default metrics is too distracting. I would remove it, and instead change `agents` to `agents/clients` in the docstring of those Metrics
- I'm not sure I would add so many metrics about number of communications (uplink, downlink, total). I would only expose the total in the codebase, without caring too much about if they are up- or downlink communications. additionally, with your suggestion on my PR, I will switch to fractional communication counters, and after addressing #312 we will support the setup where: the server transmits bigger messages than clients. the counters will then keep track of this

### default table/plot metrics
- instead of having FEDERATED_TABLE_METRICS contain DEFAULT_TABLE_METRICS, I would do concatenation `DEFAULT_TABLE_METRICS + FEDERATED_TABLE_METRICS` and maybe change the name to `FEDERATED_ONLY_TABLE_METRICS`. same goes for the other defaults
- another comments on this: we now have quite a lot of subsets of metrics to select for, and this is more confusing than helpful in my opinion (it's just more user-friendly to pass a list of metrics instead of understanding which of the defaults to use)

given these observations, the function `_resolve_default_metrics` gives me the idea: let's have only two big lists of table and plot metrics, which include all the metrics implemented in the codebase. then, we rely on the availability check to remove those that don't apply to the set-up at hand (e.g.: the federated-only ones will be unavailable for other types of networks; classification-specific metrics will not be available if the targets are not ints, so they will not be applied to regression problems); `MSE` can also be applied to classification, so there is no need to remove it, etc.

the two lists of default metrics should then be private, and the docs in `compute_metrics` would say something like: "if `None` all metrics that are available for this benchmark problem will be used. for example, federated-only metrics are removed when a P2PNetwork is passed"

nicola-bastianello:
I'm personally leaning towards option 2, but maybe there are some issues I'm not currently seeing. @adrianardv @Simpag thoughts?

adrianardv:
Thanks!

Option 1 sounds good to simplify this PR. It keeps the current metric API mostly unchanged and removes the need for separate `resulting_server_states` / `get_federated_*` . It would be enough for the current metrics.

That said, Option 2 also sounds good as a cleaner long-term design. That is, if we expect that metrics that need network-level info are added. If not, maybe option 1 is enough.

> I'm personally leaning towards option 2, but maybe there are some issues I'm not currently seeing. @adrianardv @Simpag thoughts?

Rn I can't think of any.


>- I don't see where _record_selected_clients is used and for what, maybe it can be removed?

`_record_selected_clients` is not unused. It is the helper that increments `_n_times_selected`, called from `_selected_clients_for_round`. Kept it as a helper instead of inlined cause some algorithms (like FedPD) might need to record selected clients from a slightly different flow.

> - the note "Note: For :class:~decent_bench.networks.FedNetwork, this metric is evaluated at the mean client model, not at the server model." on all default metrics is too distracting. I would remove it, and instead change agents to agents/clients in the docstring of those Metrics

Done. I've made this change.

> - I'm not sure I would add so many metrics about number of communications (uplink, downlink, total). I would only expose the total in the codebase, without caring too much about if they are up- or downlink communications. additionally, with your suggestion on my PR, I will switch to fractional communication counters, and after addressing https://github.com/team-decent/decent-bench/issues/312 we will support the setup where: the server transmits bigger messages than clients. the counters will then keep track of this

Okey, I've removed the uplink/downlink communication metrics, and keep only the TotalCommunication metric.

> default table/plot metrics
>- instead of having FEDERATED_TABLE_METRICS contain DEFAULT_TABLE_METRICS, I would do concatenation DEFAULT_TABLE_METRICS + FEDERATED_TABLE_METRICS and maybe change the name to FEDERATED_ONLY_TABLE_METRICS. same goes for the other defaults

ok

>- another comments on this: we now have quite a lot of subsets of metrics to select for, and this is more confusing than helpful in my opinion (it's just more user-friendly to pass a list of metrics instead of understanding which of the defaults to use)

> given these observations, the function _resolve_default_metrics gives me the idea: let's have only two big lists of table and plot metrics, which include all the metrics implemented in the codebase. then, we rely on the availability check to remove those that don't apply to the set-up at hand (e.g.: the federated-only ones will be unavailable for other types of networks; classification-specific metrics will not be available if the targets are not ints, so they will not be applied to regression problems); MSE can also be applied to classification, so there is no need to remove it, etc.

>the two lists of default metrics should then be private, and the docs in compute_metrics would say something like: "if None all metrics that are available for this benchmark problem will be used. for example, federated-only metrics are removed when a P2PNetwork is passed"

Nice. I like this idea. 

Okey, I've now changed it to: `compute_metrics(..., table_metrics=None, plot_metrics=None)` uses the private lists (`_DEFAULT_TABLE_METRICS` and `_DEFAULT_PLOT_METRICS`) and relies on `is_available`  to filter out the metrics tht don't apply.
Regarding the lists, instead of writing those two defaults as one long flat list, I kept the previous grouping as private lists. From the user perspective, there are no public default subsets to choose from anymore.
I can also inline everything directly into the two long private lists if you prefer.

nicola-bastianello:
> Thanks!
> 
> Option 1 sounds good to simplify this PR. It keeps the current metric API mostly unchanged and removes the need for separate `resulting_server_states` / `get_federated_*` . It would be enough for the current metrics.
> 
> That said, Option 2 also sounds good as a cleaner long-term design. That is, if we expect that metrics that need network-level info are added. If not, maybe option 1 is enough.

I'm not sure if it will be necessary, but I feel like option 2 would be better in general. however, for the scope of this PR I would go with option 2 and we can open an issue for option 2

> `_record_selected_clients` is not unused. It is the helper that increments `_n_times_selected`, called from `_selected_clients_for_round`. Kept it as a helper instead of inlined cause some algorithms (like FedPD) might need to record selected clients from a slightly different flow.

I see thank you. howere I'm a bit worried with algorithms having to explicitly call a private method like this, because users will not know to do it. and if they don't, the counting doesn't work as expected. I feel like counting needs to be taken care of entirely by the backend. is there any way that we can get rid of this function?

> Okey, I've removed the uplink/downlink communication metrics, and keep only the TotalCommunication metric.

I think after the changes according to option 1, the TotalCommunication metric will actually be redundant, since SentMessages will be the same. am I right?

> Nice. I like this idea.
> 
> Okey, I've now changed it to: `compute_metrics(..., table_metrics=None, plot_metrics=None)` uses the private lists (`_DEFAULT_TABLE_METRICS` and `_DEFAULT_PLOT_METRICS`) and relies on `is_available` to filter out the metrics tht don't apply. Regarding the lists, instead of writing those two defaults as one long flat list, I kept the previous grouping as private lists. From the user perspective, there are no public default subsets to choose from anymore. I can also inline everything directly into the two long private lists if you prefer.

great, thanks!

adrianardv:
> I see thank you. howere I'm a bit worried with algorithms having to explicitly call a private method like this, because users will not know to do it. and if they don't, the counting doesn't work as expected. I feel like counting needs to be taken care of entirely by the backend. is there any way that we can get rid of this function?

I thought about moving the coutner update into the selection schemes, but I think that would not fully solve it because algorithms may select clients without using a selection scheme. So the algorithm layer still seems to be the place where the framework actually nows which clients are selected for the round.

What I changed now is that `_selected_clients_for_round` is public: `selected_clients_for_round`. This is now the helper federated algorithms should use to select clients for each round. It applies the selection scheme, if any, and updates the selected-client counters via the private method `_record_selected_clients`.

If you see a cleaner way to make this even more automatic, let me know.

> I think after the changes according to option 1, the TotalCommunication metric will actually be redundant, since SentMessages will be the same. am I right?

If SentMessages takes the server into account and the statistics is sum, then SentMessages (sum) will in fact be the same as TotalCommunication.

However, I kept `SentMessages`, `ReceivedMessages`, and `SentMessagesDropped` client-only, as before. With that meaning, `SentMessages(avg)` remains the average across clients. If we included the server in `SentMessages`, then `SentMessages(avg)` would average together clients and the server, which feels awkward.

decent_bench/algorithms/federated/_fed_algorithm.py
```diff
@@ -62,18 +62,32 @@ def _settle_num_local_steps(self, network: FedNetwork) -> dict["Agent", int]:
             )
         return {client: self.num_local_steps[client] for client in clients}
 
-    def _selected_clients_for_round(self, network: FedNetwork, iteration: int) -> list["Agent"]:
+    def selected_clients_for_round(self, network: FedNetwork, iteration: int) -> list["Agent"]:
```
nicola-bastianello: now that the method is public, the name should be a bit better; I propose `select_clients`

decent_bench/algorithms/federated/_fed_algorithm.py
```diff
@@ -62,18 +62,32 @@ def _settle_num_local_steps(self, network: FedNetwork) -> dict["Agent", int]:
...
         """
-        Select participating clients for the current round from the currently active clients.
+        Select participating clients for the current round and record participation.
```
nicola-bastianello: I would not mention recording participation, that's something only we care about; and also remove the sentence "Calling it more than once ..."

nicola-bastianello: and change the docstring a bit more: "the method selects the subset of active clients that will receive the server broadcast and perform local training. the clients are selected using ``self.selection_scheme``"

decent_bench/algorithms/federated/_fed_algorithm.py
```diff
@@ -62,18 +62,32 @@ def _settle_num_local_steps(self, network: FedNetwork) -> dict["Agent", int]:
...
+        else:
+            selected_clients = self.selection_scheme.select(active_clients, iteration)
+        self._record_selected_clients(selected_clients)
+        return selected_clients
+
+    @staticmethod
+    def _record_selected_clients(selected_clients: Sequence["Agent"]) -> None:
+        """Record clients selected by a federated round."""
+        for client in selected_clients:
+            client._n_times_selected = getattr(client, "_n_times_selected", 0) + 1  # noqa: SLF001
```
nicola-bastianello: given that agents now always have `_n_times_selected` and `_is_server`, I don't think all the getattr guards are needed

decent_bench/algorithms/federated/_fed_pd.py
```diff
@@ -99,8 +99,10 @@ def initialize(self, network: FedNetwork) -> None:
         self._num_local_steps_by_client = self._settle_num_local_steps(network)
         self.num_local_steps = self._num_local_steps_by_client
 
-    def step(self, network: FedNetwork, _: int) -> None:
-        participating_clients = network.active_clients()
+    def step(self, network: FedNetwork, iteration: int) -> None:
+        if self.selection_scheme is not None:
```
nicola-bastianello: this is likely not the right place to do this, I think it should be done in the post_init instead

nicola-bastianello: out of curiosity: is there a technical reason why this is not possible, or is it because the original paper does not include client selection?

decent_bench/metrics/_metric.py
```diff
@@ -115,12 +115,12 @@ def get_data_from_trial(
         Get the data for this metric from a trial.
 
         Args:
-            agents: the agents being evaluated
+            agents: the snapshotted agent views being evaluated. For federated networks, this may include the server.
```
nicola-bastianello: "this may include the server" -> "this includes the server"

decent_bench/metrics/metric_library.py
```diff
@@ -7,21 +7,31 @@
...
-from decent_bench.costs import EmpiricalRiskCost
+from decent_bench.costs import Cost, EmpiricalRiskCost
 from decent_bench.metrics._metric import Metric
+from decent_bench.networks import FedNetwork
 
 if TYPE_CHECKING:
     from decent_bench.benchmark import BenchmarkProblem
 
 
+def _non_server_views(agents: Sequence[AgentMetricsView]) -> list[AgentMetricsView]:
```
nicola-bastianello: this and _server_view maybe belong in metric_utils? it could also make sense to cache them if there are no issues with AgentMetricsView not being hashable

decent_bench/metrics/metric_library.py
```diff
@@ -636,15 +653,259 @@ def get_data_from_trial(  # noqa: D102
...
         return [
-            a.cost.function(a.x_history[iteration], indices="all")
-            if isinstance(a.cost, EmpiricalRiskCost)
-            else a.cost.function(a.x_history[iteration])
-            for a in agents
+            (i, float(np.mean(self._drifts(agent_views, server, i)))) for i in utils.all_sorted_iterations([server])
         ]
 
+    @staticmethod
+    def _drifts(agents: Sequence[AgentMetricsView], server: AgentMetricsView, iteration: int) -> list[float]:
```
nicola-bastianello: this should also belong in metric_utils

decent_bench/metrics/metric_library.py
```diff
@@ -636,15 +653,259 @@ def get_data_from_trial(  # noqa: D102
...
+        server_x = server.x_history[iteration]
+        return [float(iop.norm(agent.x_history[iteration] - server_x)) for agent in agents]
+
+
+class SelectedParticipationFraction(Metric):
```
nicola-bastianello: the name is quite confusing to me and the docs can be improved; new name could be `FractionSelectedClients`; new description: "Fraction of clients selected by the federated algorithm to perform local training"

decent_bench/metrics/metric_library.py
```diff
@@ -636,15 +653,259 @@ def get_data_from_trial(  # noqa: D102
...
+    r"""
+    Fraction of client-rounds selected by the federated algorithm.
+
+    Table:
+        Average selected participation fraction over observed rounds.
```
nicola-bastianello: this can also be improved: "fraction of selected clients over the algorithm run"

decent_bench/metrics/metric_library.py
```diff
@@ -636,15 +653,259 @@ def get_data_from_trial(  # noqa: D102
...
+
+    Note:
+        Available only for :class:`~decent_bench.networks.FedNetwork`.
+
+    """
+
+    table_description: str = "selected participation fraction"
```
nicola-bastianello: fraction selected clients

decent_bench/metrics/metric_library.py
```diff
@@ -636,15 +653,259 @@ def get_data_from_trial(  # noqa: D102
...
+        __: int,
+    ) -> tuple[float]:
+        agent_views = _non_server_views(agents)
+        n_rounds = utils._observed_rounds(agent_views)  # noqa: SLF001
+        if n_rounds == 0 or not agent_views:
+            return (np.nan,)
+        return (sum(agent.n_times_selected for agent in agent_views) / (n_rounds * len(agent_views)),)
+
+
+class TotalCommunication(Metric):
```
nicola-bastianello: I understand your point, but it's too confusing to have this and also the other message counters. I would remove it and make the other counters use clients+server

nicola-bastianello (commented):
here are some comments, looks good overall

decent_bench/algorithms/federated/_fed_pd.py
adrianardv: FedPD paper does not support client selection. Both in the algorithm and the convergence analysis assume full participation. 
When communication is skipped, clients just keep training locally in the next round without aggregation + server broadcast. If we introduced client selection, the participating subset could change from round to round, so newly selected clients might be training from stale or inconsistent local/server states. This would likely affect performance. I'm guessing this is why they assume full participation in the paper.
Also, in the communication rounds, FedPD explicitly averages over all clients and broadcasts the updated server model back to all clients, which reinforces the full-participation assumption.

nicola-bastianello: I see, thanks for clarifying!

decent_bench/metrics/metric_utils.py
```diff
@@ -47,6 +47,36 @@ def _clear_caches() -> None:
     x_mean.cache_clear()
     _x_error.cache_clear()
     _predict_agent.cache_clear()
+    _non_server_views.cache_clear()
+    _server_view.cache_clear()
+
+
+def non_server_views(agents: Sequence[AgentMetricsView]) -> tuple[AgentMetricsView, ...]:
```
nicola-bastianello: I would keep `non_server_views` `server_view` and `drifts` as private (like most other utils) since the users don't care about them.

also, why do `non_server_views` `server_view` have a public and private version? please align this to the style of all other utils, with only a private util that is cached

adrianardv:
Not sure why ruff is failing now. It passes on my computer.

It seems to be comming from other files. Copilot says this:

I'll help you find a solution for the failing job. Let me start by getting the job logs.Now let me get the workflow file and the files that have the ruff errors:## Solution for Failing Job 77207283165

The ruff linter job is failing due to PLW0717 violations — try clauses containing too many statements. Ruff's default limit is 5 statements per try block, but your code has several try blocks exceeding this limit.

Affected Files and Issues
1. decent_bench/benchmark/_benchmark.py:131 (11 statements)

Lines 131-150 in the resume_benchmark function
2. decent_bench/metrics/_runtime_metric_plotter.py:65 (34 statements)

Lines 65-115 in the run method (main loop)
3. decent_bench/metrics/_runtime_metric_plotter.py:68 (19 statements)

Lines 68-92 (nested inner try block)
4. decent_bench/metrics/_runtime_metric_plotter.py:78 (8 statements)

Lines 78-92 (innermost try block)
5. decent_bench/metrics/_runtime_metric_plotter.py:243 (16 statements)

Lines 243-268 in the _process_message method
6. decent_bench/utils/checkpoint_manager.py:704 (9 statements)

Lines 704-725 in the load_metrics_result method

-------------------------------------------------------------------------------------------------
Edit: Apparently they just released ruff 0.15.14 which introduces the PLW0717 check. That's why it was suddently failling.

nicola-bastianello:
PLW0717 is an unstable rule currently in preview; and it is active because we have preview=true in the pyproject.toml. it's not a good idea to have unstable rules, so I would remove the use of rules in preview. probably removing preview=true is enough, or setting it to false. can you try that please?

while I was writing I see you have changed the code to fix this; to minimize merge conflicts with my PR, could you please roll back the last commit and try with removing preview instead?

the PR looks good otherwise

adrianardv:
I've removed preview=true. but then ruff failed on existing `#noqa` comments for preview rules. So, for now I added RUF100 to the lint.ignore list.
The alternative is to keep RUF100 enabled and remove those non-used  `#noqa` s from the codebase.
Let me know which option you prefer.

nicola-bastianello:
I think it's ok for now, we can remove unused noqa later

nicola-bastianello:
looks good thank you, merging now!

Issue #290, opened 2026-04-13
Refactor dataset loading and handling


nicola-bastianello:
Currently, the data pipeline is a bit confusing in my opinion; we have:

- `type Datapoint = tuple["Array", "Array"]`
- `type Dataset = list[Datapoint]`
- `DatasetHandler` which collects data from different sources, and can partition it; it returns a list of `Dataset`
- inside `PyTorchCost` we can also have a `torch.utils.data.DataLoader` instance
- `PyTorchDatasetHandler` accepts a pytorch `Dataset` object
- the `EmpiricalRiskCost` instances take care of sampling the data as needed

now, I see some issues:

- if one uses `PyTorchDatasetHandler` to define `PyTorchCost`, if I understand correctly: the handler loads all data, it partitions them, and each partition is passed as the `dataset` argument of a `PyTorchCost` instance. but this doesn't allow to e.g. directly pass a dataloader to `PyTorchCost`, losing the benefit of lazy loading
- in general, it is somewhat confusing to see so many dataset-related objects play different roles
- in pytorch, we have the following division of roles: 1) `Dataset` (which fetches data from some source and returns through `__getitem__`, or `__iter__` for `IterableDataset`), 2) `Sampler` which determines which data to sample from a data source, 3) `DataLoader` which wraps a `Dataset` and a `Sampler`. currently, there is no clear mapping of our objects to the pytorch data pipeline (`DatasetHandler` is kind of like a `torch.DatasetLoader`, but `Dataset` in our case is just a type, and `Sampler` is implemented directly insider the `EmpiricalRiskCost`

I think then that it would be nice to refactor all of this to the following:

1. we have a `Dataset` object which replaces the roles of both `DatasetHandler` and `Dataset` type: it can load data from whatever source in a dynamic way (with `__getitem__` or `__iter__`), exposing info about it (e.g. the length). it also exposes a `get_partitions` method which returns a list of new `Dataset` objects (instead of `list[tuple["Array", "Array"]]` like now); hopefully getting partitions can be implemented so that it doesn't have to load all data to memory
2. we expose some ready-made `Dataset` subclasses: `SyntheticClassification`, `SyntheticRegression`, `Kaggle`, `PyTorch`
3. the data definition pipeline then is something like: 1) instantiate a dataset, and partition it, 2) assign each `Dataset` into a specific subclass of `EmpiricalRiskCost`

[...]

PR #314, opened 2026-05-22
enh(datasets): Add explicit dataset partitioners

adrianardv:
This PR replaces handler-specific dataset splitting logic with explicit partitioning.

Previously, partitioning control was mostly limited to `PyTorchDatasetHandler`, where users could choose between the default random split and the `heterogeneity=True` path. Other handlers had more implicit/default behavior (random split), and the available partitioning strategies were limited.

This change separates partition generation from dataset materialization.

```python
idx_partitions = split_iid(dataset, n_partitions=10)
local_datasets = dataset.split(partitions=idx_partitions)
```

### Changes:

- Added function-based partitioning utilities:
  - `split_iid`
  - `split_stratified_iid`
  - `split_size`
  - `split_dirichlet_label`
  - `split_shard`
  - `split_label_quantity`

- Place each strategy in a separate module under datasets/partitioners.

- Replace `DatasetHandler.get_partitions()` with `DatasetHandler.split(partitions=...)`
 
- Require explicit index partitions when calling `split`; there is no implicit partitioning behavior.

- Update documentation, examples and tests.

Partitioning functions return list[list[int]], allowing users to inspect, save, modify or generate custom index partitions before materializing datasets.


Simpag:
Is this WIP or final?

nicola-bastianello:
this is a great idea, and it looks like none of the ML frameworks provide built-in tools to partition in these many ways. however, this should be addressed together with #290 later on

regarding the design: I'm not sure that class-based partition is my preferred approach. I feel like providing splitting utilities as functions might be better. something like `split_iid(dataset: Dataset, ...) -> Sequence[Dataset]`. my reasonining is that a dataset will be partitioned only once in a benchmark execution, and carrying a Partitioner instance on the Dataset might be unnecessary. the drawback with this is that the split happens in a second step instead of through `Dataset.get_partitions`

another alternative is to have `Dataset.get_partitions` (maybe renamed to `split`) to accept an argument specifying the partitioning behavior from the list of currently implemented options. the drawback with this is that we lose customizability by the user

overall I'm not opposed to using Partitioners, but I want to make sure that we don't end up complicating the API too much for very little gain, and function utilities are easier to understand in the API

nicola-bastianello:
ok, thinking some more here is my proposed idea:

- we change `Dataset.get_partitions` to receive `partitions: list[list[int],...]` which specifies a list of partitions, each as a list of the indices corresponding to the data in the partition
- we then provide utilities to generate these indices partitions; these are something like `split_X(dataset: Dataset, n_partitions, **kwargs) -> list[list[int],...]`
- we can also allow `Dataset.get_partitions` to get just the number of partitions, in which case it defaults to iid partioning: `Dataset.get_partitions(n_partitions: int | None = None,  partitions: list[list[int],...] | None = None) -> list[Dataset]`
- I would also rename `get_partitions` to `split`

when creating the datasets for the agents, we would have something like
```python
dataset = ....  # init of dataset
agents_datasets = dataset.split(partitions = split_iid(dataset, n_partitions, **kwargs))
```

adrianardv:
The separation you suggest sounds good.
If I understood correctly it would be something like:
```python
idx_partitions = split_iid(dataset, n_partitions, ...)
agents_datasets = dataset.split(partitions=idx_partitions)
```
or similarly:
```python
idx_partitions = DirichletLabelPartitioner(...).partition(n_datapoints, labels)
agents_datasets = dataset.split(partitions=idx_partitions)
```

One small detail to decide: 
For label-based splitters, the partitioning method needs to know the label of every datapoint. In the current PR the handler "hides" this detail, so the user does not have to pass labels manually.
If we move splitting outside the handler, we need to decide how `split_dirichlet_label(..)`(amongst others) gets the labels: user passes labels explicitly, handlers expose labels, ...

nicola-bastianello:
> The separation you suggest sounds good. If I understood correctly it would be something like:
> 
> ```python
> idx_partitions = split_iid(dataset, n_partitions, ...)
> agents_datasets = dataset.split(partitions=idx_partitions)
> ```

yes this is how I'm thinking of doing it

> or similarly:
> 
> ```python
> idx_partitions = DirichletLabelPartitioner(...).partition(n_datapoints, labels)
> agents_datasets = dataset.split(partitions=idx_partitions)
> ```

 I think it might be better to only provide functions to partition indices, so this would actually be
```python
idx_partitions = split_dirichlet_label(dataset, n_partitions, ...)
```

> One small detail to decide: For label-based splitters, the partitioning method needs to know the label of every datapoint. In the current PR the handler "hides" this detail, so the user does not have to pass labels manually. If we move splitting outside the handler, we need to decide how `split_dirichlet_label(..)`(amongst others) gets the labels: user passes labels explicitly, handlers expose labels, ...

if I'm not mistaken, passing the dataset to the split_X function would solve the issue, right?

adrianardv:
> if I'm not mistaken, passing the dataset to the split_X function would solve the issue, right?

Oh, yeah, that's right. Passing the dataset can solve it as long as the dataset handler exposes labels

Simpag (commented):
Move the partitioners into separate files, otherwise LGTM outside of the other comments

nicola-bastianello:
looks great, thank you! merging now

PR #315, opened 2026-05-24
enh(decent-bench): Introduce labels for messages, adapt algorithms to use labels, align arguments names across the codebase

### LINKED ISSUE #302: Support structured network messages

nicola-bastianello:
the PR introduces: 1) a refactor and extension of the agent message handling system, 2) and a consistent renaming of algorithm parameters to improve clarity and documentation. The message handling logic is now encapsulated in a new `ReceivedMessages` class, supporting labeled messages and more flexible access patterns. Argument names are now aligned throughout the codebase (e.g. `num_local_epochs` -> `num_local_steps`) and docs explicitly link mathematical symbols with argument names.

**Agent message handling improvements:**

- Introduced the `ReceivedMessages` class in `decent_bench/agents.py`, providing a structured container for received messages, supporting labels, sender-based access, and flexible clearing. This replaces the previous plain dictionary approach
- Updated agent methods to use the new `ReceivedMessages` interface, allowing retrieval of messages by label and sender, and exposing more robust APIs for message access
- Refactored message-related logic in federated algorithm base class and aggregation routines to use the new labeled message access patterns

**Argument renaming:**

- use same argument name for the same type of arguments (e.g. `penalty` not `rho`/`mu`; `relaxation` not `alpha`; `num_local_steps` not `num_local_epochs`)
- changed proximal penalty argument to `penalty` (was `rho`)
- explicitly linked arugment names to math symbols in docstrings

closes #302


decent_bench/algorithms/federated/_fed_algorithm.py
```diff
@@ -87,32 +87,44 @@ def _record_selected_clients(selected_clients: Sequence["Agent"]) -> None:
...
             client._n_times_selected += 1  # noqa: SLF001
 
-    def server_broadcast(self, network: FedNetwork, selected_clients: Sequence["Agent"]) -> None:
-        """Send the current server model to the selected clients."""
-        network.send(sender=network.server(), receiver=selected_clients, msg=network.server().x)
+    def server_broadcast(
+        self,
+        network: FedNetwork,
+        selected_clients: Sequence["Agent"],
+        label: str = "default",
```
Simpag: Maybe call it channel instead of label, makes more sense in terms of networking

decent_bench/algorithms/p2p/_kgt.py
```diff
@@ -130,8 +133,11 @@ def _update_tracking_and_params(
         # Line 10: Update model parameters
         weighted_neighbor_z = self.W[agent, agent] * z_i
         weighted_sum = self.W[agent, agent] * agent.aux_vars["msg"]
-        for j, msg in agent.messages.items():
-            weighted_neighbor_z += self.W[agent, j] * msg[1]
-            weighted_sum += self.W[agent, j] * msg[0]
+        received_state_updates = agent.messages(_STATE_LABEL)
+        received_tracking_updates = agent.messages(_GRADIENT_TRACKER_LABEL)
+        received_both = set(received_state_updates).intersection(received_tracking_updates)
+        for j in received_both:
```
Simpag, line 139: Would it be considered modifying the algorithm if you check that both arrived?

Simpag (commented):
Dont really like that channels are called labels. Otherwise there’s not much to comment on, just a question regarding the algorithms.

decent_bench/algorithms/federated/_fed_algorithm.py
nicola-bastianello: I'm partial to label, but I do agree that channel is more intuitive in the context of networks. I'll change it

decent_bench/algorithms/p2p/_kgt.py
nicola-bastianello: a bit yes, but it's not so different from only using active agents, and we also check for reception in some other algorithms. it's again the same trade-off between making algorithms as close as possible to their papers and guaranteeing that they work (that is, they don't raise) for a broad range of benchmarks

Simpag: Fair, I had a similar opinion. We should probably mention the changes either globally in the algorithms module or individually in each algorithm. Doing it in the module is probably enough

nicola-bastianello: I added a note on the user guide, when I rewrite the guide I will expand it

Simpag:
LGTM

PR #316, opened 2026-05-30
ref(metrics): Refactor the whole metrics computation and display pipeline

nicola-bastianello:
the PR refactors extensively the metrics computation and display pipeline:

### NetworkMetricsView
- introduced the `NetworkMetricsView` object; this is a `networkx.Graph[AgentMetricsView]` that preserves topology information and convenient access methods (agents/clients/server). it also exposes a list of the iterations reached by at least one agent in the network
- metrics now expect a `NetworkMetricsView` instead of a list of `AgentMetricsView`, and can directly use topology information and access methods

### Metric
- refactor to only have `compute` (which is a renaming of `get_data_from_trial`, and removed `get_table/plot_data`)
- removed statistics from Metric entirely, these are now taken care of by `compute_metrics` (via `compute_tables`)
- removed `can_diverge`, which is replaced with a logger.info message triggered by any inf/nan value in the printed table (this has the benefit that the table is neatly arranged, with only numbers)

### MetricResult
- MetricResult has been refactor to contain only the following attributes: `network_views`, `raw_table_results: dict[Metric, DataFrame]`, `raw_plot_results: dict[Metric, DataFrame]`, `table_results: DataFrame`, `plot_results: DataFrame`
- the class also exposes method `update_table_results` to recompute `table_results` with new statistics, using `raw_table_results` as data (this is just a wrapper for `compute_table_metrics`, so all logic lives there)
- it also exposes the properties `algorithms` `table_metrics` `plot_metrics` `iterations` to simplify getting a list of what is available in the data

### metrics computation
- the computation pipeline has changed to use pandas DataFrames as the data structure, and it has been reorganized
- the foundation is the new `compute_metrics_at_iter` which computes one metric on the given network_views and benchmark problem. the output is a DataFrame with columns (algorithm, trial, agent, iteration, value) which are of types (str, uint16, uint16, uint32, float32)
- the api is only `compute_metrics`, which 1) calls `compute_plot_metrics`, 2) calls `compute_table_metrics`
- `compute_plot_metrics` calls `compute_metrics_at_iter` for each metric, on all iterations, then concatenates into DataFrame with columns (algorithm, trial, agent, iteration, value)
- `compute_table_metrics` checks if the metric was already computed for plots, and in that case avoids recomputation. otherwise calls `compute_metrics_at_iter` at the last iteration. the DataFrame has comlumns (algorithm, trial, agent, value)
- to compute statistics on plots and tables, we have `aggregate_table_metrics` and `aggregate_plot_metrics`. these return a single DF each, with columns (metric, algorithm, statistic, mean, std) and (metric, algorithm, iteration, mean, min, max) respectively.
- available statistics for table metrics are: mean, std, max, min, median
- all the computations use pandas native operations
- values >1e30 or <-1e30 are replaced with +/-inf in `compute_metrics_at_iter`

### display metrics
- the structure is largely the same, with `display_metrics` the only public method, with pretty much the same signature
- `display_plots` and `display_tables` have been refactored significantly to use pandas native operations
- in `display_plots` these operations include: truncating at first occurrence of inf, replacing small values with 1e-15 or minimum positive value of the metric; shifting x if x_log=True. messaging around these operations has been clarified
- in `display_table` pandas is used to create a DataFrame with MultiIndex, which allows to then use `to_string` and `to_latex` to print the table; dependency on `tabulate` has been removed
- the logic generating the plots is largely unchanged; it changed only where it gets the data from (a DataFrame)

closes #309, closes #313


decent_bench/benchmark/_compute/compute_metrics_at_iter.py
```diff
@@ -0,0 +1,61 @@
...
+    frame.loc[
+        frame["value"].isna() | (np.isfinite(frame["value"]) & (frame["value"] > MAX_ABS_METRIC_VALUE)), "value"
+    ] = np.inf
+    frame.loc[
+        frame["value"].isna() | (np.isfinite(frame["value"]) & (frame["value"] < -MAX_ABS_METRIC_VALUE)), "value"
+    ] = -np.inf
+
+    return frame.astype(
+        {"algorithm": "category", "trial": "uint16", "agent": "uint16", "iteration": "uint32", "value": "float32"}
+    )
```
Simpag, line 61: You set value here as float 32 but above (line 50) its float64. Probably better to set all to fp32

decent_bench/benchmark/_display/_display_metrics.py
```diff
@@ -128,52 +123,61 @@ def display_metrics(  # noqa: PLR0912
...
+
+    if not selected_algorithms:
+        LOGGER.warning("No available algorithms were selected, stopping")
+        return
+    if not selected_table_metrics and not selected_plot_metrics:
+        LOGGER.warning("No available metrics were selected, stopping")
+        return
+
+    # drop metrics that were not selected
+    new_metrics_result = copy.deepcopy(metrics_result)
```
Simpag: Dont use deepcopy because agent metrics may be really large (called network metrics now), so this will use a lot of memory. Some of my experiments caused OOM due to this. I implemented the "prev_values" before to prevent this deepcopy

decent_bench/benchmark/_display/display_plots.py
```diff
@@ -0,0 +1,500 @@
...
+                    _plot_subplot(
+                        metric_subplots[metric_index_in_group],
+                        frame["iteration"].tolist(),
+                        y_mean,
+                        y_min,
+                        y_max,
+                        alg_name,
+                        alg_idx,
+                    )
+
```
Simpag: While this is fine, could reduce duplication a bit by assigning the first two arguments to `_plot_subplot` doing something like:
```python
x1 = (metric_subplots[idx*2], metric_subplots[idx*2+1]) if two_columns else (metric_subplots[idx],)
if two_columns:
  x2 = (frame["compute"], frame["iteration"])
else:
  x2 = frame["compute"] if use_cost else frame["iteration"]

for x1, x2 in zip(x1,x2, strict=True):
  _plot_subplot(x1, x2, y_mean, y_min, ...)
```

Simpag (commented):
Only major issue is the OOM using deepcopy, other than that just minor comments. Rest LGTM

Could have compute_plots and compute_tables in _compute_metrics since they are all rather short and connected.

decent_bench/benchmark/_compute/compute_metrics_at_iter.py
nicola-bastianello: good point, don't know why I did that

decent_bench/benchmark/_display/_display_metrics.py
nicola-bastianello: ah ok, that was the reason for prev_values, sorry I removed it. I'll restore the prev_values approach for metrics views, for the DFs I think I'll keep a deepcopy, unless you see issues there as well

decent_bench/benchmark/_display/display_plots.py
nicola-bastianello: nice idea!

nicola-bastianello:
> Could have compute_plots and compute_tables in _compute_metrics since they are all rather short and connected.

I prefer to keep them separate so that in the compute_plots we have compute and aggregate side by side, and it makes it easier to understand the structure of the DFs like this. everything in one file might be a bit confusing

decent_bench/benchmark/_display/_display_metrics.py
nicola-bastianello: ok, deepcopy wasn't needed for DFs as well, since the functions don't modify the original DFs. let me know if this looks good to you

Simpag: sounds good. Also, instead of doing deepcopy on dataframes, I believe that the .copy() for dataframes does a shallow copy until it deems it necessary to do a full copy. I'd probably recommend using that over deeopcopy, but might need to double check that

nicola-bastianello: yes the copy option would be the better choice. but in the new implementation not even that is needed, since the filtering operations just return a new DF (or a view of the DF)

Simpag:
LGTM

PR #318, opened 2026-06-22
ref(algorithms): make cleanup private

nicola-bastianello:
this PR makes cleanup private, and defaults to cleaning up all nodes in the networkx graph

closes #317


nicola-bastianello:
mypy issue is already fixed in #314

PR #319, opened 2026-06-23
feat(metrics): Add opt-in balanced accuracy metrics

adrianardv:
### Summary

Add opt-in balanced accuracy metrics for classification benchmarks:

- `BalancedAccuracy` for agent/client predictions
- `ServerBalancedAccuracy` for federated server predictions

These metrics are available through the metrics library but are not included in the default metric set. Users can explicitly request them when they are useful for a given experiment.

### Motivation

Plain accuracy can be misleading when classification data is highly imbalanced or heterogeneous. A model can achieve high accuracy by mostly predicting the majority class while performing poorly on minority classes.

Balanced accuracy addresses this by averaging recall across classes, giving each class equal weight regardless of its frequency. This makes it more informative for experiments where class distributions differ strongly across clients or where minority-class performance matters.

### DIFF
--- a/decent_bench/metrics/metric_library.py
+++ b/decent_bench/metrics/metric_library.py
@@ -427,6 +427,54 @@ class Accuracy(Metric):
         return utils._accuracy(network.agents(), problem, iteration)  # noqa: SLF001
 
 
+class BalancedAccuracy(Metric):
+    r"""
+    Balanced accuracy of the agents'/clients' predictions.
+
+    Table:
+        Balanced accuracy of the agents'/clients' final x.
+
+    Plot:
+        Balanced accuracy (y-axis) per iteration (x-axis).
+
+        Balanced accuracy is calculated as the mean recall obtained on each class. This makes it more informative than
+        plain accuracy for imbalanced classification datasets.
+
+    Only available for :class:`~decent_bench.costs.EmpiricalRiskCost` and integer targets.
+
+    Note:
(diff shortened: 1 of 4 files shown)


adrianardv:
mypy, sphinx/readthedocs failures have already been fixed in #314

nicola-bastianello:
thank you for this! I think it's nice to have these balanced metrics. however, we already have quite a large number of metrics, and I'm not sure we should add more to the metric_library.

what I propose is to define and use these metrics in the femnist example of #321 instead of in the main api. this way they serve as an example of "implement your own metrics", are still included, but don't complicate the api too much

what do you think?

adrianardv:
> thank you for this! I think it's nice to have these balanced metrics. however, we already have quite a large number of metrics, and I'm not sure we should add more to the metric_library.
> 
> what I propose is to define and use these metrics in the femnist example of #321 instead of in the main api. this way they serve as an example of "implement your own metrics", are still included, but don't complicate the api too much
> 
> what do you think?

I thought of including it in the metrics library because, in decentralized learning, data is often non-IID, so balanced accuracy seemed useful as an official metric.

That said, I also understand your point about keeping the metric library from growing too much. I’ll make the change now to define balanced accuracy in the FEMNIST example and use it there instead.

nicola-bastianello:
thank you for understanding. I'll close this PR and merge the other one. Thank you!

PR #320, opened 2026-06-23
enh(fednova): Simplify FedNova communication 

adrianardv:
After the introduction of channels for messages in #315, FedNova implementation in decent-bench can be simplified.

Before, two communication steps were necessary to transmit clients' normalizer and cumulative gradient. Now, that can be done in one single step, simplifying the implementation.


nicola-bastianello:
looks good, thank you!

PR #321, opened 2026-06-24
feat(examples): Add FEMNIST federated learning example

adrianardv:
## Summary
- Adds examples/FEMNIST/ as the first federated (server/client) example in decent-bench, alongside the existing peer-to-peer MNIST and NIM examples.
- src/dataset.py implements FEMNISTDatasetHandler, which loads the flwrlabs/femnist Hugging Face dataset and partitions it by writer identity, producing natural non-IID client data splits. Cites the original LEAF benchmark (Caldas et al., 2018).
- femnist.ipynb benchmarks six federated algorithms — FedAvg, FedProx, FedNova, FedAdam, FedLT, and FedDyn — with tuned hyperparameters. Ideal communication is the default; a commented-out block ets users switch to a realistic impaired-channel setting (dropout, noise, compression, partial activation).
- Updates examples/.gitignore to exclude the Hugging Face data cache (*/data/femnist).


nicola-bastianello:
looks great, thank you! I'll merge after we decide what to do with #319

PR #323, opened 2026-06-25
test: modified algorithms tests

### LINKED ISSUE #221: Add tests for distributed algorithms
Currently, there is a lack of test coverage for the distributed algorithms. Comprehensive testing is necessary to ensure correctness, robustness, and reliability of these algorithms under various scenarios. Especially useful while the framework is under heavy development where changes may cause side effects.

**Tasks:**
- Identify all distributed algorithm modules/components.
- Design unit tests for each algorithm.

**Acceptance Criteria:**
- All core distributed algorithms have comprehensive tests.
- Edge cases and fault tolerance are covered.

nicola-bastianello:
it now performs 25 iterations of each algorithm (should be enough to catch any issues); I also switched the problem to linear regression, which always finds x_optimal, so the test cannot fail due to that

closes #221

### DIFF
--- a/test/test_algorithms.py
+++ b/test/test_algorithms.py
@@ -39,52 +39,54 @@ from decent_bench.algorithms.p2p import (
     SimpleGT,
     WangElia,
 )
-from decent_bench.benchmark import create_classification_problem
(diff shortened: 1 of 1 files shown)


Simpag:
LGTM

PR #325, opened 2026-06-26
decent-bench: add version ranges to dependencies and dependabot

### LINKED ISSUE #227: Lock dependency versions
Currently we are not specifying any versions of our dependencies which is equivivalent to always updating and using the latest version. This could break features (such as the sphinx bug we saw earlier). The versions should be locked before our next release for packages that are under development and might change their API's.

nicola-bastianello:
this PR adds version ranges to all dependencies. this required some small changes around the use of pandas functions and DataFrame typing stubs that were raising some warnings.

the PR also adds the configuration for dependabot, which will take care of updating dependencies ranges with automated PRs

closes #227

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,14 +12,14 @@ classifiers = [
 ]
 license = "AGPL-3.0-only"
 dependencies = [
-  "matplotlib",
-  "networkx",
-  "numpy",
-  "pandas",
-  "rich",
-  "scipy",
-  "scikit-learn",
-  "zstandard",
+  "matplotlib>=3.9,<4",
+  "networkx>=3.4,<4",
+  "numpy>=2.0,<3",
+  "pandas>=2.3.3,<3",
+  "rich>=13,<14",
+  "scipy>=1.14,<2",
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -29,35 +29,35 @@ Issues = "https://github.com/team-decent/decent-bench/issues"
...
-  "types-networkx",
-  "torch",
-  "torchvision",
-  "types-tensorflow",
-  "pandas-stubs",
-  "kagglehub",
+  "hatch>=1.14,<2",
+  "mypy>=1.11,<2",
+  "pytest>=8,<9",
+  "ruff>=0.7,<1",
```
Simpag: I would lock ruff, mypy and sphinx versions so random updates dont break randomly, performing updates manually

Simpag (commented):
Minor comment, otherwise looks good

pyproject.toml
nicola-bastianello: good point, done. let's see if it is easy enough to maintain it like this, otherwise I will look into other options (maybe dependabot can take care of updating requirements with ==)

PR #326, opened 2026-06-26
fix: test_display was using old Metric init

nicola-bastianello:
the tests about availability of metrics in test_display were using the old Metric init. this didn't affect the tests success, but it needed updating


Simpag:
LGTM

PR #337, opened 2026-07-29
ref: reorder folders/files, remove backend tools from api reference

### LINKED ISSUE #153: Improved imports and folder structure
To aid both users and developers, we should keep imports flat (or even make them flatter) while making the folder structure more nested.

Initial requirements:
- Make the folder structure more nested (e.g. having a cost_functions subpackage and one cost function per module, putting centralized and distributed algorithms in an algorithms subpackage etc.). This helps avoid circular dependencies and makes the project easier to navigate.
- Keep imports short/flat for users' sake (e.g. users should be able to do "from decent_bench.cost_functions import QuadraticCost" rather than "from decent_bench.cost_functions.quadratic_cost import QuadraticCost").
- Hide private modules and imported symbols from users (e.g. progress bar module should be hidden from users, and if we in decent_bench.benchmark import numpy, users shouldn't be able to do, or get suggestions for, decent_bench.benchmark.numpy). 
- Give us the ability to refactor and move files without breaking the user api.

The above requirements can be achieved by making all modules and subpackages (except the user-facing top-level ones, e.g. decent-bench.metrics) private by adding a leading underscore, and reexporting public symbols in an \__all__ list in \__init__ files. 

However, more work is needed to achieve the following additional requirements:
- Sphinx generated api reference should use the short/public/reexported paths, not the private source code paths.
- Private modules should not show up as suggestions, e.g. when typing "from decent_bench.cost_functions.", the private module "_quadratic_cost" should not show up as a suggestion. This can be tricky but it is possible, for example when typing "from numpy.linalg.", the private module "_linalg" does not show up. The reasoning behind not giving users suggestions for nested modules/subpackages is that 1) we have already reexported what should be public so these suggestions are redundant and noisy, 2) they may contain private things the users shouldn't have to be aware of, 3) they export their imports.
- Checks should work for private modules (we might have to add pydocstyle as ruff doesn't check docstrings of private modules).
- Pylance should show import suggestions when typing a symbol not currently imported. This should also support nested imports (e.g. typing float64 gives the suggestion "from numpy import float64" even when numpy isn't imported yet).

Numpy has achieved many/all of these things so it could be a good starting point to look at.

nicola-bastianello:
this PR reorders the folder structure and the files (no new functionality, just reshuffling). additionally, it removes from the API reference items that are of interest only to developers

closes #153


Simpag (approved):
LGTM

